PR #4, opened 2023-10-30
Describe how MIMI protocol interacts with MIMI DS

kkohbrok:


rohan-wire (approved):
I still have issues with the text, but it is substantially better.

One suggestion, is we rename the "user" in `m.room.user` to "invite"

draft-ralston-mimi-protocol.md
```diff
@@ -181,10 +185,144 @@ by a consistent identifier.
...
+> atomically processable message.
+
+### DSEvents
+
+All events that pertain to the MLS group that underlies a room are wrapped into
+a DSEvent and sent as an Event of type `ds.group`.
+
+~~~ tls
+struct {
+   DSRequest ds_request;
```
turt2live, line 262: From an API design perspective, ideally we split `DSRequest` out to individual events.

draft-ralston-mimi-protocol.md
```diff
@@ -181,10 +185,144 @@ by a consistent identifier.
...
+struct {
+  opaque room_id;
+  UserData participants<V>;
+  // TODO: Add any remaining room data
+} RoomState;
+~~~
+
+Whenever a client creates a commit, it includes an UpdateRoom proposal that
+contains all events since the preceeding commit and updates the ParticipantData
+extension accordingly as part of the commit.
```
turt2live, line 237: All events sounds like a lot. Do we just need specific event types here? `m.room.user` primarily.

turt2live (commented):
from a first pass of review

draft-ralston-mimi-protocol.md
kkohbrok: Not a strong opinion, but I thought this was a good way of maintaining the abstraction of a cryptographic building block.

draft-ralston-mimi-protocol.md
kkohbrok: It should probably be "all events that change the room state", which currently is only m.room.user.

PR #5, opened 2023-11-01
events via MLS proposals

kkohbrok:


draft-ralston-mimi-protocol.md
```diff
@@ -227,30 +247,39 @@ struct {
...
-  UserEvent user_event<V>;
-  // TODO: Other events that change room state
-} UpdateRoom;
-~~~
+Whenever a client creates a commit, it MUST include all valid proposals accepted
+by the Hub during the current epoch. This includes both proposals that carry
+room-state changes, as well as proposals sent as part of MIMI DS events.
+
+Note that the validity of a proposal depend on the current room state, which may
+change during anepoch based on room-state changing events. The changes of these
```
turt2live, line 264:
```suggestion
change during an epoch based on room-state changing events. The changes of these
```

draft-ralston-mimi-protocol.md
```diff
@@ -227,30 +247,39 @@ struct {
...
+Certificate of the Hub.
 
-> **TODO**: Client should be able to staple events to a commit to create an
-> atomically processable message.
+When a user from a follower server becomes a participant in the room, the
+Certificate of the follower server MAY be added to the extension. When the last
+participant belonging to a follower server leaves the room, the certificate of
+that user MUST be removed from the list. Changes to the `external_senders`
+extension only take effect when the MLS proposal containing the event is
+committed by a MIMI DS commit. See {{ev-mroomuser}} for more information.
```
turt2live, line 282: Is this to provide the follower an ability to propose removal of their own users' clients? It feels like a path towards hub portability, which I'm very happy to include, though not sure if we want to open that can of worms yet.

draft-ralston-mimi-protocol.md
```diff
@@ -576,25 +604,47 @@ the room. They do not have access to the cryptographic state.
 
 **Event type**: `m.room.user`
 
-**Additional event fields**:
+An `m.room.user` event can be used to change the participation state of a room's
+participant.
+
+> **TODO**: Do we also want this to be able to change a participant's role?
```
turt2live, line 610: Role and participation are very different concepts imo. They could be part of the atomic operations though, where we add a user, their clients, and promote them to admin in a single transaction/frame. 

Definitely worth leaving as a TODO for now 😇

draft-ralston-mimi-protocol.md
```diff
@@ -576,25 +604,47 @@ the room. They do not have access to the cryptographic state.
 
 **Event type**: `m.room.user`
 
-**Additional event fields**:
+An `m.room.user` event can be used to change the participation state of a room's
+participant.
```
turt2live, line 608:
```suggestion
An `m.room.user` event can be used to change the participation state of a user.
```

to avoid "participation" appearing twice.

draft-ralston-mimi-protocol.md
```diff
@@ -576,25 +604,47 @@ the room. They do not have access to the cryptographic state.
...
+> encapsulate more than just this event.
 
 ~~~
 struct {
    // The user ID being affected by this participation state change.
    opaque targetUserId;
 
    // The new participation state for the target user.
-   ParticipationState state;
+   Participan state;
```
turt2live, line 634: accidental edit?
```suggestion
   ParticipationState state;
```

turt2live (approved):
Overall this is looking good. We should probably spin up a plausible DoubleRatchet implementation of this document to ensure compatibility, but that doesn't need to happen for working group purposes.

draft-ralston-mimi-protocol.md
kkohbrok: Kind of. It's for the follower server to be able to send signaling messages at all (or rather for everyone else to be able to authenticate them). This is a limitation stemming from the way that proposals from external senders (i.e. senders who are not group members) are authenticated. My thinking was that if a user is part of a room, its server might want to be able to send signaling messages independent of the user's clients. If it wants to do that, it needs to be added to the list. This has to happen before the server can send any signaling messages (or at least those implemented by MLS proposals) and it can only happen via commit by a client.

kkohbrok (commented):
Thanks for the review and the suggestions! I'll leave this open for a bit longer before merging, but I think it's decent enough to start building the next PR on top of it.

draft-ralston-mimi-protocol.md
```diff
@@ -113,15 +113,16 @@ list, a cryptographic representation of the room as defined by
...
-clients.
+performed exclusively with the hub server for a room.
+
+Events that change the state of the room are implemented through MLS proposals
+as defined in {{!RFC9420}}, thus allowing the underlying MIMI DS protocol to
+anchor the current room state cryptographically. MLS proposals are signed,
+allowing every recipient in their path to verify their authenticity.
+
+This document defines additional events to encapsulate MIMI DS protocol
+messages, for example, to add clients to a room's underlying MLS group or to
```
rohan-wire, line 124:
```suggestion
messages, for example, to
```

draft-ralston-mimi-protocol.md
```diff
@@ -159,12 +160,8 @@ struct {
...
    select (Event.type) {
       case "m.room.user":
-         // Some event types, like m.room.user, require a signature because
-         // the event transits over multiple network hops. The signature
-         // covers the whole Event struct, and is generated by the origin
-         // server for the Event.
-         Signature signature;
-         UserEvent content; // see later in doc
+         // MLSMessage containing a UserEvent proposal
+         MLSMessage user_event_proposal; // see later in doc
```
rohan-wire, line 164:
```suggestion
         Proposal user_event_proposal; // see later in doc
```

draft-ralston-mimi-protocol.md
```diff
@@ -178,45 +175,68 @@ struct {
...
-> **TODO**: We might want to move this section elsewhere.
+As soon as the Hub accepts an event that changes the room state, its effect is
+applied to the room state and future events are validated in the context of that
+new state.
+
+The room state is thus changed based on events, even if the MLS proposal
+implementing the event was not yet committed by a client. Note that this only
+applies to events changing the room state, but not for MIMI DS specific events
+that change the group state. For more information on the proposal-commit
+paradigm and the role of the MIMI DS protocol see {{mimi-ds}}.
```
rohan-wire, line 208: This is still very awkward. I'd just add a TODO to reword it and I can work on some proposed text later.

draft-ralston-mimi-protocol.md
```diff
@@ -178,45 +175,68 @@ struct {
...
+
+
+## Cryptographic room representation {#mimi-ds}
 
 Each room is represented cryptographically by an MLS group and the Hub that
-manages the room acts as the DS as defined in the MIMI DS protocol specified in
-{{!I-D.robert-mimi-delivery-service}}.
+manages the room acts as the delivery service (DS) as defined in the MIMI DS
```
rohan-wire, line 214: Isn't the DS supposed to be spread across all the clients and all the providers?  Maybe something like the "the arbiter of the delivery service (DS)"?

Feel free to ignore this comment until later.

rohan-wire (approved):

draft-ralston-mimi-protocol.md
kkohbrok: Fair point. I'll clarify that in the next PR.

PR #10, opened 2023-11-02
Adapt flows to previous changes

kkohbrok:


draft-ralston-mimi-protocol.md
```diff
@@ -491,10 +491,19 @@ described in {{anchoring}}.
...
-An *invite* is when a user (or more specifically, a user's client) is attempting
-to introduce *all* of another user's clients to the room and cryptographic state.
-This is first done by updating the target user's participation state through the hub
-server for the room.
+> **TODO**: For now, the invite flow implies that the user has to explicitly
+> accept by adding one or more clients via external commit as part of the "Join"
+> flow. In the future, we can either make the "Invite" flow more versatile and
+> allow for Welcome based invitations, or create an additional "Add" flow, that
+> allows participants to add other users (and their clients) directly via
+> Welcome.
```
rohan-wire, line 499: This is the opposite of the normal flow in most messengers. We need an Add flow and it should be first.

draft-ralston-mimi-protocol.md
```diff
@@ -718,6 +751,9 @@ sent to additional servers, as required by {{invites}}, {{joins}}, {{leaves}},
 
 ## `m.room.participant_list` {#ev-mroomparticipant_list}
 
+> **TODO**: We probably won't need this event, as `ds.fetch_group_info` contains
```
rohan-wire, line 754: Just delete the section please

rohan-wire (changes requested):

draft-ralston-mimi-protocol.md
rohan-wire: The user visible behavior is that Alice adds Bob and immediately sends a message. The expectation in most messengers is that Bob's clients can view the message that Alice just sent.

kkohbrok: I think there are two distinct flows here that I think we should cover.

1. The "invite" flow, where Alice invites Bob and Bob only becomes part of the room once he has accepted the invitation, and
2. the "add" flow, where Alice adds Bob and Bob immediately becomes a part of the room without any additional involvement by Bob.

Flow 1 would be based on external commit and corresponds, for example, to Signal's flow with invite links. Flow 2 is the flow you allude to, based on Welcome, and is based on Signal's flow, where once Alice has a connection with Bob, she can add him to other groups without Bob's involvement.

kkohbrok: The issue I see with the "Add" flow is that we probably want to add some degree of consent signaling, s.t. the Hub and other participants can verify that Alice has Bob's permission to Add Bob to the room.

rohanmahy (approved):

draft-ralston-mimi-protocol.md
rohanmahy: looks good

rohan-wire (approved):

PR #12, opened 2023-11-03
Generalize framing and add Response struct

kkohbrok:


draft-ralston-mimi-protocol.md
```diff
@@ -190,6 +213,46 @@ by a consistent identifier.
...
+   // Additional fields may be present as dependent on event type.
+   select (EventResponse.type) {
+      case ok:
+         struct {};
+      case key_package:
+         DSResponse key_package;
+      case group_info:
+         DSResponse group_info;
+      case error:
+         EventError error;
```
rohan-wire, line 251: In general, I think we need to handle "provisional" messages that are sent to the hub, and fanned out messages differently. The meaning of an error response to a fanout message means "something happened and I can't deal with this fanout message now" but the hub is at that point fully "committed" and could reach other clients. Whereas a message en route to the hub could be rejected and won't make it to any clients.

rohan-wire (approved):

PR #20, opened 2023-11-06
Remove extraneous m. in front of room events

rohan-wire:
globally replace `m.room.` with `room.` in event names.

### DIFF
--- a/draft-ralston-mimi-protocol.md
+++ b/draft-ralston-mimi-protocol.md
@@ -241,14 +241,14 @@ it becomes the Hub for the room.
 ## Alice adds Bob
 
 Management of the participants list (i.e. the list of users in the room) is done
-through `m.room.user` events. To add Bob to the room, Alice creates an
-`m.room.user` event. See {ev-mroomuser} for more information on user state
+through `room.user` events. To add Bob to the room, Alice creates an
+`room.user` event. See {ev-mroomuser} for more information on user state
 changes.
 
 Room state is anchored in the room's underlying MLS group through a GroupContext
 Extension, which mirrors all of the room's state variables.
 
-`m.room.user` events are MLS proposals, which change the room state outside of
+`room.user` events are MLS proposals, which change the room state outside of
 the MLS group immediately upon reception and align the mirror of the room state
 in the extension (which is part of the group state) upon the next `ds.commit`
 event. This reflects MLS' proposal-commit paradigm that allows members and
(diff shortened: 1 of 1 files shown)


draft-ralston-mimi-protocol.md
turt2live, line 1: The prefix exists to ensure namespacing is possible, as folks may want to add their own event types to the protocol. For example, to accomplish DoubleRatchet use cases. 

It was my intention to also prefix the DS events with `m.` for consistency and building the official namespace. The namespace needs to be codified in the IANA registry still, but that was largely assumed to be a later problem.

`m.` is shorthand for `mimi.`, which we can expand if needed.

turt2live (changes requested):
see threaded comment

bifurcation:
Yeah, I'm inclined to leave this for the WG to comment on.  The traditional IETF approach to something like this would be to have an IANA registry where certain values are registered (those that would be `m.` here) and certain values would be vendor space (I believe the Matrix pattern for these is `com.example.foo`).  If we were talking about numerical code points, the analogy would be exact, since if you know the range of numbers reserved for vendor usage (≈ non-`m.` here), you can look at a value and tell if it's registered or vendor.

For text-like registries, things are a little less clear.  Traditionally, these registries don't have an explicit vendor reservation, since space is not constrained.  (See, e.g., the [ACME directory metadata fields registry](https://www.iana.org/assignments/acme/acme.xhtml#acme-directory-metadata-fields).)  If you want a value for some proprietary thing, you just choose one and use it, and if some official thing later collides with it, oh well.  And in cases where there were explicit affordances for experimental (e.g., `X-` and `P-` in HTTP/SIP headers), those affordances [are now deprecated](https://datatracker.ietf.org/doc/html/rfc6648).

So it's not immediately clear to me what the right answer is here, and how to do extensibility seems like a good question for the WG to weigh in on.

draft-ralston-mimi-protocol.md
rohan-wire: I propose we use `private.` and a reverse domain name label for namespacing of private events. 
MIMI  events then can use room for room events, ds for ds events, etc.

rohan-wire: yes

rohan-wire: and if we use an enum managed by IANA then we can use `room` and `dr` but the `dr` constant might be initially in the private range.

rohan-wire: If we use an enum instead for the IANA registry then we get the best of both worlds

rohan-wire:
Closing as this is superfluous in PR #50

PR #39, opened 2023-11-13
Remove outdated TODO

### LINKED ISSUE #37: Refactor invites?
> **TODO**: For now, the invite flow implies that the user has to explicitly
> accept by adding one or more clients via external commit as part of the "Join"
> flow. In the future, we can either make the "Invite" flow more versatile and
> allow for Welcome based invitations, or create an additional "Add" flow, that
> allows participants to add other users (and their clients) directly via
> Welcome.

kkohbrok:
Closes #37.

We already have an Add and an Invite flow, which I think solves this TODO.


turt2live (approved):
thanks!

PR #42, opened 2024-01-29
Update Introduction and add Known Gaps

bifurcation:
The Known Gaps section might need some updating depending on what we do with the rest of the document.

### DIFF
--- a/draft-ralston-mimi-protocol.md
+++ b/draft-ralston-mimi-protocol.md
@@ -69,7 +69,7 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
 
 # Introduction
 
-The More Instant Messaging Interoperability (MIMI) protocol enables providers of
+The More Instant Messaging Interoperability (MIMI) transport protocol enables providers of
 end-to-end encrypted instant messaging to interoperate. As described in the MIMI
 architecture {{?I-D.barnes-mimi-arch}}, group chats and direct messages are
 described in terms of "rooms".  Each MIMI protocol room is hosted at a single
@@ -77,7 +77,7 @@ provider (the "hub" provider"), but allows users from different providers to
 become participants in the room.  The hub provider maintains the policy and
 participation for the room, authorizes messages, and is responsible for message
 ordering, the participation list, and policy of its rooms. Each provider also
-stores initial keying material and consent for its users (who may be offline).
+stores initial keying material and consent for its own users (who may be offline).
 
 This document describes the communication among different providers necessary to
 support messaging application functionality, for example:
@@ -96,14 +96,14 @@ Layer Security (MLS) protocol {{!RFC9420}}, and each room is associated with an
 MLS group. MLS also ensures that clients in a room agree on the room policy and
 participation.  MLS is integrated into MIMI in such a way as to ensure that a
 client is joined to a room's MLS group only if the client's user is a
-participant in the group, and that all clients in the group agree on the state
+participant in the room, and that all clients in the group agree on the state
 of the room (including, for example, the room's participant list).
 
 ## Known Gaps
 
 In this version of the document, we have tried to capture enough concrete
 functionality to enable basic application functionality, while defining enough
-of a protocol framework to indicate how other necessary functionality.  The
+of a protocol framework to indicate how to add other necessary functionality.  The
 following functions are likely to be needed by the complete protocol, but are
 not covered here:
 
--- a/draft-ralston-mimi-protocol.md
+++ b/draft-ralston-mimi-protocol.md
@@ -88,9 +88,8 @@ support messaging application functionality, for example:
 * Exchanging secure messages
 
 In support of these functions, the protocol also has primitives to fetch initial
-keying material, fetch the current state of the underlying end-to-end encryption
-protocol for the room, and request, grant, and reject consent to communicate
-with users on other providers.
+keying material and fetch the current state of the underlying end-to-end encryption
+protocol for the room.
 
 Messages sent inside each room are end-to-end encrypted using the Messaging
 Layer Security (MLS) protocol {{!RFC9420}}, and each room is associated with an
--- a/draft-ralston-mimi-protocol.md
+++ b/draft-ralston-mimi-protocol.md
@@ -121,12 +121,11 @@ support.
 Consent:
 : In this document, we assume that any required consent has already been
 obtained, e.g., a user consenting to be added to a room by another user.  The
-full protocol will need some mechanisms for establishing this consent.  Some
-initial notes are in {{consent}}
+full protocol will need some mechanisms for establishing this consent.
 
 Identifiers:
 : Certain entities in the MIMI system need to be identified in the protocol.  In
-{{identifiers}}, we define a notional syntax for identifiers, but a more
+this document, we define a notional syntax for identifiers, but a more
 concrete one should be defined.
 
 Abuse reporting:
--- a/draft-ralston-mimi-protocol.md
+++ b/draft-ralston-mimi-protocol.md
@@ -138,6 +138,8 @@ might be different from the identifier that should be used internally.  For
 example, a user-visible handle might need to be mapped to a durable internal
 identifier.  This document provides no mechanism for such resolution.
 
+Authentication
+: While MLS provides basic message authentication, users should also be able to (cryptographically) tie the identity of other users to their respective providers. Further authentication such as tying clients to their users (or the user's other clients) may also be desirable.
 
 # Conventions and Definitions
 
--- a/draft-ralston-mimi-protocol.md
+++ b/draft-ralston-mimi-protocol.md
@@ -74,9 +74,10 @@ end-to-end encrypted instant messaging to interoperate. As described in the MIMI
 architecture {{?I-D.barnes-mimi-arch}}, group chats and direct messages are
 described in terms of "rooms".  Each MIMI protocol room is hosted at a single
 provider (the "hub" provider"), but allows users from different providers to
-become participants in the room.  The hub provider maintains the policy and
-participation for the room, authorizes messages, and is responsible for message
-ordering, the participation list, and policy of its rooms. Each provider also
+become participants in the room. The hub provider is responsible for ordering
+and distributing messages, enforcing policy, and authorizing messages. It also
+keeps a copy of the room state, which includes the room policy and participant
+list, which it can provide to new joiners. Each provider also
 stores initial keying material and consent for its own users (who may be offline).
 
 This document describes the communication among different providers necessary to
--- a/draft-ralston-mimi-protocol.md
+++ b/draft-ralston-mimi-protocol.md
@@ -78,7 +78,7 @@ become participants in the room. The hub provider is responsible for ordering
 and distributing messages, enforcing policy, and authorizing messages. It also
 keeps a copy of the room state, which includes the room policy and participant
 list, which it can provide to new joiners. Each provider also
-stores initial keying material and consent for its own users (who may be offline).
+stores initial keying material for its own users (who may be offline).
 
 This document describes the communication among different providers necessary to
 support messaging application functionality, for example:
--- a/draft-ralston-mimi-protocol.md
+++ b/draft-ralston-mimi-protocol.md
@@ -69,27 +69,75 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
 
 # Introduction
 
-Providers of secure messaging services typically allow their users to only
-communicate with other users of the same provider. This can be limiting for
-users, who frequently have to run clients of multiple messaging applications in
-parallel, each for a different secure messaging provider.
+The More Instant Messaging Interoperability (MIMI) protocol enables providers of
+end-to-end encrypted instant messaging to interoperate. As described in the MIMI
+architecture {{?I-D.barnes-mimi-arch}}, group chats and direct messages are
+described in terms of "rooms".  Each MIMI protocol room is hosted at a single
+provider (the "hub" provider"), but allows users from different providers to
+become participants in the room.  The hub provider maintains the policy and
+participation for the room, authorizes messages, and is responsible for message
+ordering, the participation list, and policy of its rooms. Each provider also
+stores initial keying material and consent for its users (who may be offline).
 
-The More Instant Messaging Interoperability (MIMI) protocol specified in this
-document allows interoperability on the provider level based on rooms. Each MIMI
-protocol room is hosted at a single provider, but allows users from different
-providers to become participants in the room.
+This document describes the communication among different providers necessary to
+support messaging application functionality, for example:
 
-Users and providers can use events to interact with the provider hosting the
-room and with one-another, for example, to add or remove users from the room.
+* Sharing room policy
+* Adding and removing participants in a room
+* Exchanging secure messages
 
-Each user can have clients in the room, who can then communicate with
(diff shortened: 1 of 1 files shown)


draft-ralston-mimi-protocol.md
```diff
@@ -69,27 +69,74 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
+participation.  MLS is integrated into MIMI in such a way as to ensure that a
+client is joined to a room's MLS group only if the client's user is a
+participant in the group, and that all clients in the group agree on the state
+of the room (including, for example, the room's participant list).
+
+## Known Gaps
+
+In this version of the document, we have tried to capture enough concrete
+functionality to enable basic application functionality, while defining enough
+of a protocol framework to indicate how other necessary functionality.  The
```
rohanmahy, line 106:
```suggestion
of a protocol framework to indicate how to add other necessary functionality.  The
```

draft-ralston-mimi-protocol.md
```diff
@@ -69,27 +69,74 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
-user and client level membership.
+The More Instant Messaging Interoperability (MIMI) protocol enables providers of
+end-to-end encrypted instant messaging to interoperate. As described in the MIMI
+architecture {{?I-D.barnes-mimi-arch}}, group chats and direct messages are
+described in terms of "rooms".  Each MIMI protocol room is hosted at a single
+provider (the "hub" provider"), but allows users from different providers to
+become participants in the room.  The hub provider maintains the policy and
+participation for the room, authorizes messages, and is responsible for message
+ordering, the participation list, and policy of its rooms. Each provider also
+stores initial keying material and consent for its users (who may be offline).
```
rohanmahy, line 80:
```suggestion
stores initial keying material and consent for its own users (who may be offline).
```

rohanmahy (approved):

draft-ralston-mimi-protocol.md
```diff
@@ -69,27 +69,74 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
-and ensures that clients in the room agree on room state and group state such as
-user and client level membership.
+The More Instant Messaging Interoperability (MIMI) protocol enables providers of
+end-to-end encrypted instant messaging to interoperate. As described in the MIMI
+architecture {{?I-D.barnes-mimi-arch}}, group chats and direct messages are
+described in terms of "rooms".  Each MIMI protocol room is hosted at a single
+provider (the "hub" provider"), but allows users from different providers to
+become participants in the room.  The hub provider maintains the policy and
+participation for the room, authorizes messages, and is responsible for message
+ordering, the participation list, and policy of its rooms. Each provider also
```
kkohbrok, line 79: Is there a reason we're distinguishing between "maintains the policy and participation for the room" and then later in the sentence "and is responsible for (...) the participation list, and policy of its rooms"?

draft-ralston-mimi-protocol.md
kkohbrok, line 80: We should be more clear here what it means to "store consent". At least it's not clear to me.

draft-ralston-mimi-protocol.md
```diff
@@ -69,27 +69,74 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
+stores initial keying material and consent for its users (who may be offline).
+
+This document describes the communication among different providers necessary to
+support messaging application functionality, for example:
+
+* Sharing room policy
```
kkohbrok, line 85: What does it mean to "share room policy"? Agreement between providers? Or participants?

In any case, the other two examples seem more natural in the context of "messaging application functionality", so I'd put them first.

draft-ralston-mimi-protocol.md
```diff
@@ -69,27 +69,74 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
+
+Abuse reporting:
+: There is no mechanism in this document for reporting abusive behavior to a
+messaging provider.
+
+Identifier resolution:
+: In some cases, the identifier used to initiate communications with a user
+might be different from the identifier that should be used internally.  For
+example, a user-visible handle might need to be mapped to a durable internal
+identifier.  This document provides no mechanism for such resolution.
```
kkohbrok, line 139: If we do want to keep a list (I think another PR removes existing remarks about missing things), we should add authentication here. We do have basic MLS authentication, but we should also be able to verify cryptographically, that a given user belongs to a given provider.

kkohbrok (commented):
Thanks for the PR. Made an initial pass with a few questions and remarks.

draft-ralston-mimi-protocol.md
rohanmahy: I meant maintains the policy and participation as *enforcing* the policy and participation. It's also the source of truth ("responsible") for the "message ordering, the participation list, and policy of its rooms".

Feel free free to propose some alternate wording.

draft-ralston-mimi-protocol.md
rohanmahy: if a user has granted, revoked, or rejected consent, that needs to get stored at their provider so the provider can act on behalf of the user if his or clients are offline.

draft-ralston-mimi-protocol.md
rohanmahy: I'd say "share room policy" is between participants and the hub.

draft-ralston-mimi-protocol.md
```diff
@@ -69,27 +69,74 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
-
-Each user can have clients in the room, who can then communicate with
-one-another, where messages are end-to-end encrypted using the Messaging Layer
-Security (MLS) protocol ({{!RFC9420}}).
-
-An MLS group (through the MIMI DS protocol
-{{!I-D.robert-mimi-delivery-service}}) forms the cryptographic core of each room
-and ensures that clients in the room agree on room state and group state such as
-user and client level membership.
+The More Instant Messaging Interoperability (MIMI) protocol enables providers of
```
turt2live, line 72:
```suggestion
The More Instant Messaging Interoperability (MIMI) transport protocol enables providers of
```

draft-ralston-mimi-protocol.md
turt2live: in my opinion, we should descope consent for now. It needs a lot more thought as to how it works.

```suggestion
stores initial keying material for its own users (who may be offline).
```

draft-ralston-mimi-protocol.md
```diff
@@ -69,27 +69,74 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
+support messaging application functionality, for example:
+
+* Sharing room policy
+* Adding and removing participants in a room
+* Exchanging secure messages
+
+In support of these functions, the protocol also has primitives to fetch initial
+keying material, fetch the current state of the underlying end-to-end encryption
+protocol for the room, and request, grant, and reject consent to communicate
+with users on other providers.
```
turt2live, line 92: descoping consent: 
```suggestion
In support of these functions, the protocol also has primitives to fetch initial
keying material and fetch the current state of the underlying end-to-end encryption
protocol for the room.
```

draft-ralston-mimi-protocol.md
```diff
@@ -69,27 +69,74 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
+
+Messages sent inside each room are end-to-end encrypted using the Messaging
+Layer Security (MLS) protocol {{!RFC9420}}, and each room is associated with an
+MLS group. MLS also ensures that clients in a room agree on the room policy and
+participation.  MLS is integrated into MIMI in such a way as to ensure that a
+client is joined to a room's MLS group only if the client's user is a
+participant in the group, and that all clients in the group agree on the state
```
turt2live, line 99:
```suggestion
participant in the room, and that all clients in the group agree on the state
```

draft-ralston-mimi-protocol.md
```diff
@@ -69,27 +69,74 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
+Authorization policy:
+: In this document, we assume that all participants in a room have equal
+capability.  Actual messaging systems have authorization policies for which
+clients can take which actions in a room.
+
+Advanced join/leave flows:
+: In this document, all adds / removes / joins / leaves are initiated from
+within the group, since this aligns well with MLS.  Messaging application
+support a variety of other flows, some of which this protocol will need to
+support.
```
turt2live, line 119: This is a bit too vague: what flows specifically are we descoping from this version?

turt2live (changes requested):
Mentions of consent outside the known gaps section should be elided for now.

draft-ralston-mimi-protocol.md
rohan-wire: This statement merely says that the providers are responsible for keeping track of the consent of their users, which should be non-controversial. This section / PR doesn't impose any specific mechanism for doing so.

draft-ralston-mimi-protocol.md
rohan-wire: Richard wanted to make a join flow easy to understand by the WG, get consensus to make it a WG item, then add other join mechanisms. So in this version I think Richard wants to start with the "member adds non-member" flow. 

I'm fine (keen even) to add a second flow where a user adds a new client or a user already on a participant list (passively) does an external join, but only if we have strong consensus from the design team. 

@turt2live are you OK with that plan of action?

draft-ralston-mimi-protocol.md
kkohbrok: Thanks for clarifying. I'd be okay to leave it as is for now and touch it again once we know more about how consent works.

draft-ralston-mimi-protocol.md
```diff
@@ -69,27 +69,74 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
-and ensures that clients in the room agree on room state and group state such as
-user and client level membership.
+The More Instant Messaging Interoperability (MIMI) transport protocol enables providers of
+end-to-end encrypted instant messaging to interoperate. As described in the MIMI
+architecture {{?I-D.barnes-mimi-arch}}, group chats and direct messages are
+described in terms of "rooms".  Each MIMI protocol room is hosted at a single
+provider (the "hub" provider"), but allows users from different providers to
+become participants in the room.  The hub provider maintains the policy and
+participation for the room, authorizes messages, and is responsible for message
+ordering, the participation list, and policy of its rooms. Each provider also
```
kkohbrok, line 79: Something like this? 
```suggestion
become participants in the room.  The hub provider is responsible for message ordering and distribution, as well as policy enforcement. It also keeps a copy of the room state, which includes the room's policy and participant list, which it can provide to new joiners. Each provider also
```
I wouldn't say it's the "source of truth", since the truth is in the MLS group state and the MLS commits. Saying it's the source of truth (to me) implies that it has direct control over it. It stores the group state, but the group state ultimately comes from the clients (even though the hub can reject commits it doesn't like).

draft-ralston-mimi-protocol.md
kkohbrok: It still sounds a bit weird to me, but I don't want to nitpick too much here. Let's leave it as is for now.

draft-ralston-mimi-protocol.md
```diff
@@ -69,27 +69,74 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
+
+Abuse reporting:
+: There is no mechanism in this document for reporting abusive behavior to a
+messaging provider.
+
+Identifier resolution:
+: In some cases, the identifier used to initiate communications with a user
+might be different from the identifier that should be used internally.  For
+example, a user-visible handle might need to be mapped to a durable internal
+identifier.  This document provides no mechanism for such resolution.
```
kkohbrok, line 140: Maybe something like this?
```suggestion
Authentication
: While MLS provides basic message authentication, users should also be able to (cryptographically) tie the identity of other users to their respective providers. Further authentication such as tying clients to their users (or the user's other clients) may also be desirable.
```

draft-ralston-mimi-protocol.md
```diff
@@ -69,27 +69,74 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
```
rohan-wire, line 139:
```suggestion
identifier.  This document provides no mechanism for such resolution.

```

draft-ralston-mimi-protocol.md
```diff
@@ -69,28 +69,77 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
-user and client level membership.
-
+The More Instant Messaging Interoperability (MIMI) transport protocol enables providers of
+end-to-end encrypted instant messaging to interoperate. As described in the MIMI
+architecture {{?I-D.barnes-mimi-arch}}, group chats and direct messages are
+described in terms of "rooms".  Each MIMI protocol room is hosted at a single
+provider (the "hub" provider"), but allows users from different providers to
+become participants in the room.  The hub provider maintains the policy and
+participation for the room, authorizes messages, and is responsible for message
+ordering, the participation list, and policy of its rooms. Each provider also
```
rohan-wire, line 79:
```suggestion
become participants in the room. The hub provider is responsible for ordering
and distributing messages, enforcing policy, and authorizing messages. It also
keeps a copy of the room state, which includes the room policy and participant
list, which it can provide to new joiners. Each provider also
```

kkohbrok (approved):
Thanks for making the changes! Looks good to me now.

draft-ralston-mimi-protocol.md
kkohbrok: The text looks good now, except that I don't quite get what the difference between "enforcing policy" and "authorizing messages" is. My understanding was that authorizing messages (or not) is part of policy enforcement.

draft-ralston-mimi-protocol.md
bifurcation: Yeah, we're descoping everything besides the basic flow from this version.  The idea here is to capture the basic moving parts for one use cases, with the idea that the WG can then elaborate them to cover other use cases after adoption.

draft-ralston-mimi-protocol.md
```diff
@@ -69,28 +69,77 @@ MIMI protocol relies on MLS ({{!RFC9420}}, through the MIMI DS protocol
...
+The More Instant Messaging Interoperability (MIMI) transport protocol enables providers of
+end-to-end encrypted instant messaging to interoperate. As described in the MIMI
+architecture {{?I-D.barnes-mimi-arch}}, group chats and direct messages are
+described in terms of "rooms".  Each MIMI protocol room is hosted at a single
+provider (the "hub" provider"), but allows users from different providers to
+become participants in the room. The hub provider is responsible for ordering
+and distributing messages, enforcing policy, and authorizing messages. It also
+keeps a copy of the room state, which includes the room policy and participant
+list, which it can provide to new joiners. Each provider also
+stores initial keying material and consent for its own users (who may be offline).
```
bifurcation, line 81:
```suggestion
stores initial keying material for its own users (who may be offline).
```

turt2live (commented):
several days late (sorry, FOSDEM), but lgtm - thanks for applying the edits :)

PR #45, opened 2024-01-29
Add the MLS AppSync mechanism

bifurcation:


draft-ralston-mimi-protocol.md
```diff
@@ -1221,6 +1221,146 @@ The hub server SHOULD consider a network error as a rejection. It is expected
...
+
+[[ NOTE: This design exposes the high-level structure of the application state
+to MLS.  An alternative design would be to have the application state be opaque
+to MLS.  There is a trade-off between generality and the complexity of the API
+between the MLS implementation and the application.  An opaque design would give
+the application more freedom, but require the MLS stack to call out to the
+application to get the updated state as part of Commit processing.  This design
+allows the updates to happen within the MLS stack, so that no callback is
+needed, at the cost of forcing the application state to fit a certain structure.
+]]
```
rohanmahy, line 1256:
```suggestion
> **NOTE:** This design exposes the high-level structure of the application state
> to MLS.  An alternative design would be to have the application state be opaque
> to MLS.  There is a trade-off between generality and the complexity of the API
> between the MLS implementation and the application.  An opaque design would give
> the application more freedom, but require the MLS stack to call out to the
> application to get the updated state as part of Commit processing.  This design
> allows the updates to happen within the MLS stack, so that no callback is
> needed, at the cost of forcing the application state to fit a certain structure.
> It also potentially can result in smaller state updates in large groups.
```

draft-ralston-mimi-protocol.md
```diff
@@ -1221,6 +1221,146 @@ The hub server SHOULD consider a network error as a rejection. It is expected
...
+The state for Each `applicationId` in the `application_states` needs to conform
+to one of four basic types: an ordered array, an unordered array, a map, or an
+irreducible blob. This allows the AppSync proposal to efficiently modify a large
+application state object.
+
+The content of the `application_states` extension and the `AppSync` proposal are
+structured as follows:
+
+~~~ tls
+struct {
```
rohanmahy, line 1267:
```suggestion
~~~ tls
enum {
    irreducible(0),
    map(1),
    unorderedList(2),
    orderedArray(3),
    (255)
} StateType;

struct {
```

draft-ralston-mimi-protocol.md
```diff
@@ -1221,6 +1221,146 @@ The hub server SHOULD consider a network error as a rejection. It is expected
...
+Note that the `application_states` extension is updated directly by AppSync
+proposals; a GroupContextExtensions proposal is not necessary. A proposal list
+that contains both an AppSync proposal and a GroupContextExtensions proposal
+is invalid.
+
+Likewise a proposal list in a Commit MAY contain more than one AppSync proposal,
+but no more than one AppSync proposal per `applicationId`. The proposals are
+applied in the order that they are sent in the Commit.
+
+AppSync proposals do not need to contain an UpdatePath. An AppSync proposal can be sent by an authorized external sender.
```
rohanmahy, line 1359:
```suggestion
AppSync proposals do not need to contain an UpdatePath. An AppSync proposal can
be sent by an authorized external sender.
```

rohanmahy (approved):

kkohbrok (commented):
Do I understand correctly, that there can be only one ApplicationState per appliction_id? In that case, each application can only have one map/array/etc.? What if the application needs different data structures or multiple of one kind?

More generally, I'm a bit concerned about the introduction of the application_id since it's not part of MLS negotiation mechanism. It's not relevant to MLS interop, because the semantics of the application are outside of MLS, but it seems relevant that clients can signal support for a given application, or that a group can require that an application be supported. I guess what I'm trying to say is that either there is agreement between clients on the application that runs on top of MLS in which case we don't need an application_id, or there isn't, in which case we need a negotiation mechanism.

Regarding opaque vs. non-opaque state, I think I have a slight preference for letting the application handle the AppSync proposals and modify the state itself. The current design seems a bit clunky and limiting.

rohan-wire:
> Do I understand correctly, that there can be only one ApplicationState per appliction_id? In that case, each application can only have one map/array/etc.? What if the application needs different data structures or multiple of one kind?

Each application_id contains one top-level map/array/etc. An application can have multiple application_ids (each representing a specific type of application state). In barnes-mimi-protocol we proposed two application_ids (participant list, and room base policy). If an application _id needs a complex structure, the top level data structure for each application_id could be a map (which can contain arbitrarily complex data structures below it) or it could be an irreducible blob.

> More generally, I'm a bit concerned about the introduction of the application_id since it's not part of MLS negotiation mechanism. It's not relevant to MLS interop, because the semantics of the application are outside of MLS, but it seems relevant that clients can signal support for a given application, or that a group can require that an application be supported. I guess what I'm trying to say is that either there is agreement between clients on the application that runs on top of MLS in which case we don't need an application_id, or there isn't, in which case we need a negotiation mechanism.

I would say the later, but do we need that negotiation in MLS? I was thinking that existence of specific application IDs in the GroupContext means that that state does exist in the group. I think adding a new application ID to the GroupContext would require agreement in a GroupContextExtension proposal or out of band, but I think that negotiation is ok at the app level. I could be convinced otherwise however.

> 
> Regarding opaque vs. non-opaque state, I think I have a slight preference for letting the application handle the AppSync proposals and modify the state itself. The current design seems a bit clunky and limiting.

I can totally accept that considering I put it together in about 45 minutes. My main motivation was to allow a patch to one element of the participant list or one setting of many to be a small patch instead of, for example, sending the entire participant list of 1000 users just to add or remove one. Anything that accomplishes that goal, or a demonstration that the bandwidth/processing is a non-issue, would make me more comfortable with the opaque blob model.

draft-ralston-mimi-protocol.md
```diff
@@ -1221,6 +1221,146 @@ The hub server SHOULD consider a network error as a rejection. It is expected
...
+
+[[ TODO: IANA registry for `application_id`; register extension and proposal types
+as safe extensions ]]
```
turt2live, line 1362: Should this get the same quoted TODO treatment as the others?

turt2live (approved):
lgtm, though I'm not an MLS expert. Some of the system design characteristics might change over time, I suspect.

draft-ralston-mimi-protocol.md
```diff
@@ -1221,6 +1221,146 @@ The hub server SHOULD consider a network error as a rejection. It is expected
 that the original sender will attempt to re-send the invite once the server is
 reachable again.
 
+# MLS Application State Synchronization
```
turt2live, line 1224: Clarifying question: can the server see the contents of the application state?

rohan-wire: The hub sees everything, all the servers see all the changes sent in AppState proposals or in GroupContextExtension proposals. We haven't specified if the complete application state needs to be conveyed to other providers at the time a Welcome is received.

kkohbrok:
Thanks for the clarifications!

I'm not sure it's a great idea on the one hand for applications to require an application_id per piece of state they want to store and on the other hand have application_id be IANA registered. My suggestion would be that each application use an extension id and then define its own AppSync mechanism. This allows the application to define an arbitrary number of proposals with arbitrary semantics, as well as arbitrary state to store in the group context.

draft-ralston-mimi-protocol.md
rohan-wire, line 1362:
```suggestion
> **TODO:** IANA registry for `application_id`; register extension and proposal types
as safe extensions
```

rohan-wire:
> Thanks for the clarifications!
> 
> I'm not sure it's a great idea on the one hand for applications to require an application_id per piece of state they want to store and on the other hand have application_id be IANA registered. My suggestion would be that each application use an extension id and then define its own AppSync mechanism. This allows the application to define an arbitrary number of proposals with arbitrary semantics, as well as arbitrary state to store in the group context.

Applications can still define arbitrary semantics of their state, they just don't get arbitrary patching semantics in AppSync. If the thing that you want to change is irreducible, replace the whole thing. If the thing you want to change is a map, list, or array, you can replace/delete/insert into those oft-used data structures without resending the entire (possibly large) data structure. That's all. If folks only want "replace the whole thing" semantics, I can live with that, but I am definitely not ok with having the patch semantics vary per application. I've seen that protocol movie before and it bombed.

PR #47, opened 2024-02-06
Add sections introducing layering

rohan-wire:


draft-ralston-mimi-protocol.md
```diff
@@ -459,6 +459,103 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+source and target providers for the request, in the following way:
+
+* The target provider is indicated using a Host header {{!RFC9110}}.  If the
+  provider is using a non-standard port, then the port component of the Host
+  header is ignored.
+* The source provider is indicated using a From header {{!RFC9110}}.  The
+  `mailbox` production in the From header MUST use the `addr-spec` variant, and
+  the `local-part` of the address MUST contain the fixed string `mimi`.  Thus,
+  the content of the From header will be `mimi@a.example`, where `a.example` is
+  the domain name of the source provider.
```
turt2live, line 476: The `From` header I think needs further discussion with the working group - it feels like abuse of the header to me. Ideally we don't try to wedge ourselves into a header if that header isn't capable of clearly and semantically representing our use case - a new header or mechanism may be required, which is fine imo.

draft-ralston-mimi-protocol.md
```diff
@@ -459,6 +459,103 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+endpoints defined in {{application-layer}}.
+
+## End-to-End Security Layer
+
+Every MIMI room has an MLS group associated to it, which provides end-to-end
+security guarantees.  The clients participating in the room manage the MLS-level
+membership by sending Commit messages covering Add and Remove proposals.
+
+Every message sent within a room is authenticated and confidentiality-protected
+by virtue of being encapsulated in an MLS PrivateMessage object.
```
turt2live, line 492: Every message meaning user/instant message, or MLS Message? (please clarify in text)

draft-ralston-mimi-protocol.md
```diff
@@ -459,6 +459,103 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+The PublicMessage encapsulation provides sender authentication, including the
+ability for actors outside the group (e.g., servers involved in the room) to
+originate AppSync proposals.  Encoding room state changes in MLS proposals
+ensures that a client will not process a commit that confirms a state change
+before processing the state change itself.
+
+[[ TODO: A little more needs to be said here about how MLS is used.  For
+example: What types of credential are required / allowed?  If servers are going
+to be allowed to introduce room changes, how are their keys provisioned as
+external signers? ]]
```
turt2live, line 509: should this be converted to the quoted comment style we seem to have adopted elsewhere?

draft-ralston-mimi-protocol.md
```diff
@@ -459,6 +459,103 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+the room's policy. Non-hub servers facilitate interactions between their clients
+and the hub server.
+
+In this section, we describe the state that servers keep. The following top
+level section describes the HTTP endpoints exposed to enable these functions.
+
+### Server State
+
+Every MIMI server is a publication point for users' key material, via the
+`keyMaterial` endpoint discussed in fetch-key-material.  To support this
```
turt2live, line 526: `fetch-key-material` looks like it's supposed to be a link?

turt2live (commented):
generally lgtm - mostly clarifying comments contained within. Thanks!

draft-ralston-mimi-protocol.md
rohan-wire: it was, but it relies on PR #48

draft-ralston-mimi-protocol.md
rohan-wire, line 492:
```suggestion
Every application message sent within a room is authenticated and confidentiality-protected
by virtue of being encapsulated in an MLS PrivateMessage object.
```

draft-ralston-mimi-protocol.md
rohan-wire, line 509:
```suggestion
> **TODO**: A little more needs to be said here about how MLS is used.  For
example: What types of credential are required / allowed?  If servers are going
to be allowed to introduce room changes, how are their keys provisioned as
external signers?
```

draft-ralston-mimi-protocol.md
bifurcation:
```suggestion
`keyMaterial` endpoint discussed in fetch-key-material [TODO: link].  To support this
```

draft-ralston-mimi-protocol.md
bifurcation:
```suggestion
* The source provider is indicated using a From header {{!RFC9110}}.  The
  `mailbox` production in the From header MUST use the `addr-spec` variant, and
  the `local-part` of the address MUST contain the fixed string `mimi`.  Thus,
  the content of the From header will be `mimi@a.example`, where `a.example` is
  the domain name of the source provider.
  
[[ NOTE: The use of the From header field here is not really well-aligned with its
intended use.  The WG should consider whether this is correct, or whether a new
header field would be better.  Perhaps something like "From-Host" to match Host? ]]
```

bifurcation (approved):

PR #48, opened 2024-02-06
Incorporate the basic mechanism described in the example flows.

rohan-wire:


draft-ralston-mimi-protocol.md
```diff
@@ -459,894 +459,420 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+~~~
+POST /keyMaterial/{targetUser}
+~~~
 
-The MIMI DS protocol uses MLS, which follows a proposal-commit paradigm. Any
-party involved in a room (follower server, Hub or clients) can send proposals
-(e.g. to add/remove/update clients of a user or to re-initialize the group with
-different parameters). However, only clients can send commits, which contain all
-valid previously sent proposals and apply them to the MLS group state.
+The target user's URI is listed in the request path. KeyPackages
```
turt2live, line 514:
```suggestion
The target user's identifier is listed in the request path. KeyPackages
```

draft-ralston-mimi-protocol.md
```diff
@@ -459,894 +459,420 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+The request body has the following form.
 
 ~~~ tls
 struct {
-   opaque user_id;
-   opaque role;
-   ParticipationState state;
-} ParticipantData
+    opaque uri<V>;
+} IdentifierUri;
```
turt2live, line 534: I think we ideally generify the identifier shape for now, at least until the working group has identifier discussions (and more specifically, URI shape/use).

For now we can probably just call this `Identifier` with an `opaque id<V>; // TODO: Define shape` property

draft-ralston-mimi-protocol.md
```diff
@@ -459,894 +459,420 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+    IdentifierUri userUri;
+    ClientKeyMaterial clients<V>;
+} KeyMaterialResponse;
 ~~~
 
-**Additional validation rules**:
+At minimum, as each MLS KeyPackage is returned to a requesting provider (on
+behalf of a requesting IM client), the target provider needs to associate its
+`KeyPackageRef` with the target client and the hub provider needs to associate
+its `KeyPackageRef` with the target provider. This insures that Welcome messages
```
turt2live, line 617:
```suggestion
its `KeyPackageRef` with the target provider. This ensures that Welcome messages
```

draft-ralston-mimi-protocol.md
```diff
@@ -459,894 +459,420 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
-
-**Event type**: `m.room.info`
-
-**Additional event fields**:
+~~~ tls
+enum {
+  success(0),
+  wrongEpoch(1),
+  notAllowed(2),
+  hubUnresponsive(3),
```
turt2live, line 699: why would the hub be unresponsive? this isn't a client-server endpoint, so there's no feedback to provide with respect to the error here.

draft-ralston-mimi-protocol.md
```diff
@@ -459,894 +459,420 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
-  |           Async fanout of event | Async fanout of event        |
-  |<--------------------------------|----------------------------->|
-  |                                 |                              |
+~~~ tls
+struct {
+  Protocol protocol;
+  select(protocol) {
+    case mls10:
+      /* PrivateMessage containing an application message */
+      MLSMessage app_message;
```
turt2live, line 727: consistency:

```suggestion
      MLSMessage appMessage;
```

draft-ralston-mimi-protocol.md
```diff
@@ -459,894 +459,420 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
-> **TODO**: Include validation rules for permissions.
-
-**Fanout considerations**:
-
-Each `m.room.user` event is fanned out as normal ({{fanout}}). The event MAY be
-sent to additional servers, as required by {{invites}}, {{joins}}, {{leaves}},
-{{bans}}, {{knocks}}.
+> **ISSUE:** Do we want to offer a distinction between regular application
+messages and ephemeral applications messages (for example "is typing"
+notifications), which do not need to be queued at the target provider.
```
turt2live, line 757: (yes, I think we do, but that's a different PR/issue's concern imo)

draft-ralston-mimi-protocol.md
```diff
@@ -459,894 +459,420 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
 
-**Additional validation rules**:
+## Fanout Messages and Room Events
 
-None.
+If the hub provider accepts an application or handshake message (proposal or
+commit) message, it forwards that message to all other providers with active
+participants in the room and all local clients which are active participants.
```
turt2live, line 763:
```suggestion
participants in the room and all local clients which are active members.
```

draft-ralston-mimi-protocol.md
```diff
@@ -459,894 +459,420 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
-{{membership}}.
+The state of a room consists of its room ID, its base policy, its
+participant list (including the role and participation state of each
+participant), and the associated end-to-end protocol state (its MLS group
+state) that anchors the room state cryptographically.
 
-Additional servers MAY have the event sent to them if required by the steps
-leading up to fanout.
+While all parties involved in a room agree on the room's state during a
+specific epoch, the Hub is the arbiter that decides if a state change is valid.
```
turt2live, line 824: There should something to this effect somewhere: "the Hub MUST NOT accept changes which are non-compliant with the room's policy, and similarly MUST NOT decline changes which are compliant with the room's policy"

This is to prevent arbitrary rules from being created/enforced by the hub.

draft-ralston-mimi-protocol.md
```diff
@@ -459,894 +459,420 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
-ensuring the invite does not appear spammy in nature and if the inviter already
-has a connection with the invitee.
+## Cryptographic room representation
 
-If the server does not recognize the event format of the `InfoEvent`
-({{ev-mroomcreate}}) event, or does not understand the policy/encryption
-configuration contained within, it MUST reject the request.
+Each room is represented cryptographically by an MLS group and the Hub that
+manages the room uses the MIMI DS protocol specified in
+{{!I-D.robert-mimi-delivery-service}} to manage that group.
```
turt2live, line 840: sorry, it took me until this line to realize that the diff is really weird and broken. Would you be able to structure the changes so the diff is usable?

turt2live (changes requested):
This is a partial review based on the diff presented. It looks like something went very sideways with the diff though - would you be able to restructure the changes to make the diff parsable please?

rohan-wire:
> This is a partial review based on the diff presented. It looks like something went very sideways with the diff though - would you be able to restructure the changes to make the diff parsable please?

If we had support for stacked diffs this would be easier. I suppose I could redo a new equivalent PR with individual commits: one commit deleting, one commit adding, then one or more commits modifying things that reference the new content. I find it hard to get a clean diff from git for large markdown changes. Git way too aggressively looks for one common word dozens of lines in the middle of a replacement.

If you have any concrete suggestions @turt2live @bifurcation please let me know.

draft-ralston-mimi-protocol.md
rohan-wire: Is there any reason _NOT_ to use a URI?  A URI allows us to encapsulate existing and new identifiers without collisions, and we can trivially include a URI into certificates and into MLS data structures. I think this first property is absolutely essential. I also can't think of any other solution which gives us that property and can be included in X.509 certificates which already work with RFC5280.

To be clear, there is no need to agree on a specific URI scheme or schemes.

draft-ralston-mimi-protocol.md
rohan-wire: many of these provider interfaces will be decomposed, especially in a high availability environment. This would be the case if the ingress is functioning and the specific service or even the database was not working.  Think of this as analogous to a 503 HTTP response.

draft-ralston-mimi-protocol.md
rohan-wire: @turt2live 
You were nearly though the PR by this point. Maybe you would be able to review the remaining 5 paragraphs and bullet list changes to the Acknowledgements? Perhaps using the rich diff mode?

draft-ralston-mimi-protocol.md
rohan-wire: > There should something to this effect somewhere: "the Hub MUST NOT accept changes which are non-compliant with the room's policy, and similarly MUST NOT decline changes which are compliant with the room's policy"
> 
> This is to prevent arbitrary rules from being created/enforced by the hub.

I think normative language like this should be in the `/update` endpoint section. 

I would be wary of the "MUST NOT decline" for things like rate limiting and spam/abuse prevention. As well, the hub may have sent a valid external AppSync proposal which crossed an otherwise valid commit on the wire. I think it should be up to the hub whether it a) accepts the valid commit and resends its proposal, or b) rejects the commit with suitable information (incorporate proposal ref xyz)  and have the provider's client retry.

draft-ralston-mimi-protocol.md
turt2live: There are certainly good reasons to use a URI, but I don't feel we've sufficiently discussed the concept to consider it having consensus yet. In the interest of progress, we should remove the mention here and talk about it as a working group post-adoption.

draft-ralston-mimi-protocol.md
turt2live: why can't we use HTTP 503 then? I'd be a bit surprised if folks configured their reverse proxies to return MIMI-specific error codes in this case.

draft-ralston-mimi-protocol.md
turt2live: This section sounds normative too - should it be in both?

There's certainly wording adjustments to be made to account for rate limiting, of course, though I don't think we should be allowing the hub to declare something is spam to get out of a policy obligation (for example). For the AppSync example, this sounds like a problem for sequencing: if the commit comes through before the proposal exists (even if the server knows it's generating that proposal), then it's a legitimate commit that *must* be accepted. The next commit will be required to incorporate the proposal. If the proposal is appended first though, the commit will be required to incorporate it at that time. This is all dependent on whether the protocol document actually requires the hub server's external proposals to be incorporated though - this is not something that should be an implementation detail.

draft-ralston-mimi-protocol.md
turt2live: The side-by-side diff unfortunately doesn't really help either. I usually approach these sorts of diffs with multiple passes, but I'm finding it incredibly hard to do those passes over this PR, sorry.

turt2live:
It appears as though Git is getting confused with which text is "first", which commonly happens when new text is added above modified text. It may help to put the net-new sections below the modified/deleted sections of the existing text.

draft-ralston-mimi-protocol.md
rohan-wire: > There are certainly good reasons to use a URI, but I don't feel we've sufficiently discussed the concept to consider it having consensus yet. In the interest of progress, we should remove the mention here and talk about it as a working group post-adoption.

We had strong guidance from the to make our proposal concrete. Nobody has proposed an alternative identifier as good as a URI or any reason not to use one, so in the interest of concreteness, I'd like to leave it a URI.

draft-ralston-mimi-protocol.md
rohan-wire: > why can't we use HTTP 503 then? I'd be a bit surprised if folks configured their reverse proxies to return MIMI-specific error codes in this case.

Sure.

draft-ralston-mimi-protocol.md
rohan-wire: > This section sounds normative too - should it be in both?
> 
> There's certainly wording adjustments to be made to account for rate limiting, of course, though I don't think we should be allowing the hub to declare something is spam to get out of a policy obligation (for example). For the AppSync example, this sounds like a problem for sequencing: if the commit comes through before the proposal exists (even if the server knows it's generating that proposal), then it's a legitimate commit that _must_ be accepted. The next commit will be required to incorporate the proposal. If the proposal is appended first though, the commit will be required to incorporate it at that time. This is all dependent on whether the protocol document actually requires the hub server's external proposals to be incorporated though - this is not something that should be an implementation detail.

The hub decides what arrives first. From the hub's perspective, its external proposal was sent and fanned out and on the wire. It receives the commit later, so that commit according to the ordering policy of mimi is invalid because it does not include all valid pending proposals. The same thing could happen if Alice sends a commit to add Carol and Bob sends a proposal to remove himself. If Bob's proposal arrives first *from the hub's perspective*, Alice's Commit is invalid.

If the hub sends an otherwise valid proposal from a valid `external_sender`, the proposal needs to be incorporated. The policy remedy if a group/room doesn't want that is to configure the group without an `external_sender` for the hub and/or only join rooms without an `external_sender` configured in the corresponding room.

draft-ralston-mimi-protocol.md
turt2live: There hasn't been a proposal because, to my knowledge, it's not been considered by people yet. If we have a strong proposal to use URIs, that can remain as a PR and discussed at IETF 119, but not part of the document itself for now.

draft-ralston-mimi-protocol.md
turt2live: I'm failing to see how that contradicts my original suggested wording, sorry.

rohan-wire:
> It appears as though Git is getting confused with which text is "first", which commonly happens when new text is added above modified text. It may help to put the net-new sections below the modified/deleted sections of the existing text.

Replaced with PR #50 . I'll close this one in favor of the other.

PR #49, opened 2024-02-06
Longer security considerations section

rohan-wire:


draft-ralston-mimi-protocol.md
```diff
@@ -1499,10 +1499,26 @@ be sent by an authorized external sender.
...
+verified by the end clients involved in the room.  The usage of MLS ensures that
+the servers facilitating the exchange cannot read messages in the room or
+falsify room state changes, even though they can read the room state change
+messages.
+
+Each room has an authorization policy that dictates which protocol actors can
+perform which actions in the room.  This policy is enforced by the hub server
+for the room.  The actors for whom the policy is being evaluated authenticate
+their identities to the hub server using the MLS PublicMessage signed object
+format, together with the identity credentials presented in MLS.
```
turt2live, line 1521: we may need to add a note that the hub-and-spoke model allows the hub to be a threat actor itself, but is deemed appropriate for simplicity.

turt2live (commented):
otherwise lgtm

draft-ralston-mimi-protocol.md
rohan-wire: good point

bifurcation:
```suggestion
Each room has an authorization policy that dictates which protocol actors can
perform which actions in the room.  This policy is enforced by the hub server
for the room.  The actors for whom the policy is being evaluated authenticate
their identities to the hub server using the MLS PublicMessage signed object
format, together with the identity credentials presented in MLS.  This design
means that the hub is trusted to correctly enforce the room's policy, but this 
cost is offset by the simplicity of not having multiple policy enforcement points.
```

bifurcation (approved):

PR #50, opened 2024-02-09
Incorporate the basic mechanism described in the example flows (take 2)

rohan-wire:
As requested by Travis, this is a replacement for PR #48 which has cleaner diffs (via individual commits).


draft-ralston-mimi-protocol.md
```diff
@@ -1651,18 +1160,15 @@ Example: `m.room.info`
...
 
-* {{?I-D.ralston-mimi-signaling}} describes {{event-schema}},
-  {{room-creation}}, details of {{membership}}, and subsections of {{rest-api}}.
+* {{?I-D.ralston-mimi-policy}},
 
-Aspects of {{?I-D.ralston-mimi-policy}} are additionally taken into
-consideration in this document through subsections of {{membership}}, but is
-largely unincorporated and may require updates to match this document's
-specifics.
+* {{?I-D.mahy-mimi-group-chat}},
```
turt2live, line 1171: `ralston-policy` and `mahy-group-chat` are *not* consolidated into this document. The only portion of `ralston-policy` which was incorporated was the notion of participation state, which seems to have gone missing under this PR - that needs to be reinstated in some capacity.

I'd suggest reinstating the "Aspects of ralston-policy..." paragraph, and either clarify how `mahy-group-chat` has had an impact on this document or remove it.

draft-ralston-mimi-protocol.md
turt2live, line 1: I think we also need an example showing how 2 servers negotiate the encryption being used for a room, and how that applies when a third server is trying to become involved.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
+`leaf_node.lifetime.not_after` time has passed.
 
-* Clients can only be added to the group if the associated user is on the
-  participant list and in the `join` state.
 
-#### Fetch KeyPackage {#ev-fetchkeypackage}
+## Update Room State
 
-**Event type**: `ds.fetch_key_packages`
+Adds, removes, and policy changes to the room are all forms of updating the
```
turt2live, line 726: policy changes or policy *configuration*? It shouldn't be possible to change the policy envelope.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
-struct {
-  DSRequest fetch_group_info;
-} DSFetchGroupInfo
+POST /update/{roomId}
 ~~~
 
-**Additional validation rules**:
+In MLS 1.0, any change to the room base policy document or to the participant
+list is always communicated via an `AppSync` proposal type with the appropriate
+`applicationId`. The participant list change needed to add a user MUST
```
turt2live, line 737: what is the appropriate `applicationId`?

turt2live (dismissed):
This is a first pass of the PR, trying to cover as much of the content as possible. I'll aim to do further passes over the next few days and week.

See comments contained within for details. There's a mix of editorial comments and functional changes.

draft-ralston-mimi-protocol.md
rohan-wire: Why isn't configuration sufficient?  We are probably talking about a handful of manually configured providers per hub.

draft-ralston-mimi-protocol.md
rohan-wire: the appropriate applicationId for the base policy document is the MIMI base policy application ID. The appropriate applicationId for the participation list is the MIMI participation list application ID.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
-ensuring the invite does not appear spammy in nature and if the inviter already
-has a connection with the invitee.
+## Cryptographic room representation
 
-If the server does not recognize the event format of the `InfoEvent`
-({{ev-mroomcreate}}) event, or does not understand the policy/encryption
-configuration contained within, it MUST reject the request.
+Each room is represented cryptographically by an MLS group and the Hub that
+manages the room uses the MIMI DS protocol specified in
+{{!I-D.robert-mimi-delivery-service}} to manage that group.
```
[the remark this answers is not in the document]
rohan-wire: I don't think that is very clear. Does the client that talks to the hub as a local provider need to implement MIMI DS, or just use compatible MLS objects and signatures?  For example, the hub needs some way to get all components of the commit bundle, but the specific struct used to convey those components doesn't have to be the same.

draft-ralston-mimi-protocol.md
rohan-wire: How about "Concepts from the following documents motivated the contents of this document..."?

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
turt2live: I'm not following, sorry. Clients aren't providers - they belong to providers. What information is missing in this PR which ensures MLS is reliable for users of providers?

draft-ralston-mimi-protocol.md
turt2live: I'm not sure either of the policy documents have really provided enough inspiration to be listed, honestly. The participation state enum was defined well before the policy documents existed, and should be returned to this PR somewhere (I can work on wording, if helpful). 

What parts of `mahy-group-chat` were incorporated?

draft-ralston-mimi-protocol.md
turt2live: I think that needs to be made clearer in the document. I was under the impression that we're trying to connect providers regardless of the DMA legal framework, where any random messaging provider could approach another MIMI-compatible provider on the open internet.

draft-ralston-mimi-protocol.md
turt2live: That's configuration of the policy, not fundamentally changing the policy rules which exist.

draft-ralston-mimi-protocol.md
turt2live: I'm not following, sorry. Can we get verbiage in the document please?

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
rohan-wire: What I am saying is that there are two things being described by MIMI DS. One is a list of expected behaviors related to MLS, including many of things described in the "Operational Considerations" section of the MLS architecture document. This is largely aligned. The second is almost an API to access the KeyPackages, handle Commits, access GroupInfo, etc. A provider that implemented what is in this PR, could then use an "API" very similar to the endpoints in MIMI DS to accomplish what the MIMI provider needs to implement the related to MLS actions. 

There are three main differences:
- this document assumes that the MIMI provider knows about users (which could include pseudonymous users) and can determine if a given client "belongs to" that user. 
- MIMI DS requires the hub to process the whole ratchet tree just like a client (which could be a scalability concern for large providers), whereas this PR has the `RatchetTreeRepresentation` assuming the simple thing first (the client uploads the ratchet tree).
- MIMI DS has certain design decisions based on enabling metadata privacy. we need to tease that out more (which Raphael, Konrad, and I have already committed to do), but for now we want to show the basic case work

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
 
-MIMI protocol messages are sent described using the TLS presentation language
-format ({{Section 3 of RFC8446}}).
+This section describes the specific endpoints necessary to provide the
+functionality in the example flow. The framing for each endpoint includes a
+protocol so that different variations of the end-to-end encryption can be used.
 
-All MIMI protocol messages are framed by a MIMIMessage.
+The syntax of the MIMI protocol messages are described using the TLS
+presentation language format ({{Section 3 of RFC8446}}).
```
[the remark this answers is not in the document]
rohan-wire: I think that is well understood by the working group at this point. This point has been discussed at 3 IETFs now.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
+unable to forward a Welcome message to the target provider).
 
-The MIMI DS protocol ensures that the Hub, all follower servers and the clients
-of all participants (or at least those in the `join` state) agree on the group
-state, which includes the client list and the key material used for message
-encryption (although the latter is only available to clients). Since the group
-state also includes a copy of the room state at the time of the most recent
-commit, it is also covered by the agreement.
+The path includes the target user. The request body includes the protocol
+(currently just MLS 1.0), and the requesting user. The request SHOULD include
```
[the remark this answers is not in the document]
rohan-wire: Let's leave these stylistic editorial improvements for another PR once this merges please. I think it is great to have a PR that adds internal references, but only after a merge. I already have a TODO link in another place.

draft-ralston-mimi-protocol.md
rohan-wire: I think we are just acknowledging other work that came before us. Describing exactly what came from where or was even ultimately included sounds like a non-goal at this stage.

draft-ralston-mimi-protocol.md
rohan-wire: We agreed to leave out things that don't motivate the basic Alice and Bob flow. In that flow, both sides assume they can use MLS as described here.

draft-ralston-mimi-protocol.md
rohan-wire: perhaps you could provide a concrete example of this distinction?

draft-ralston-mimi-protocol.md
rohan-wire: Sure, once this is merged and we have the AppSync extension expanded, I can writeup an IANA section that registers the applicationIds in new PR. That should make it super clear.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
 `update` endpoint on the hub.
 
-# Framing
+# MIMI Endpoints and Framing
 
-MIMI protocol messages are sent described using the TLS presentation language
-format ({{Section 3 of RFC8446}}).
+This section describes the specific endpoints necessary to provide the
+functionality in the example flow. The framing for each endpoint includes a
+protocol so that different variations of the end-to-end encryption can be used.
```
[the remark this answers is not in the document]
turt2live: A note would be fine, yea.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
turt2live: understanding should be documented, imo.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
turt2live: Including those differences in the document itself would really help understanding, I think. I'm still not fully following the ratchet tree stuff, but will give it another read this week.

draft-ralston-mimi-protocol.md
turt2live: If we aren't incorporating the work, I don't think we need to reference it here. The working group will need to have discussions about policy as a separate discussion/document, which can and should reference both drafts.

I'd suggest removing both `ralston-policy` and `mahy-group-chat` from the acknowledgements for now.

draft-ralston-mimi-protocol.md
turt2live: The envelope is the set of rules which form an authorization policy itself. The room may have toggles to steer behaviour, but how those flags work is set in stone through the envelope.

For example, the permissions system might define 2 flags:
* `can_ban` - Whether users belonging to the role can ban other users.
* `can_invite` - Whether users belonging to the role can invite other users.

The envelope could have a rule which says a user must be unbanned before they can be invited, which implies behaviour upon users which hold `can_invite`. All participating providers have a shared understanding for this envelope, and can locally enforce those operations without having to disturb the hub with needless requests. The actual assignment of the flags is left as a configuration detail for the room, though does still need to be visible to the providers which are locally enforcing policy (for obvious cases that don't require the hub).

draft-ralston-mimi-protocol.md
turt2live: This PR needs a note about the application IDs in it. The IANA section can be expanded in a future PR, but for this PR to be mergeable the description of what the application ID look like needs to be included.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
bifurcation: I'm not going to go to the mat on this, but having one endpoint handle multiple versions actually seems quite messy to me.

Since we have a directory now, we can do versioning there, e.g., by having both `notify` and `notify_v2` endpoints.  Then the actual endpoints only have to handle one version.

draft-ralston-mimi-protocol.md
```diff
@@ -459,6 +459,357 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+
+~~~ tls
+enum {
+    reserved(0),
+    mls10(1),
+    (255)
+} Protocol;
+~~~
+
+## Directory
```
bifurcation, line 580: One thing I notice we lost in the deleted sections was some prose exempting the `.well-known` endpoint from the mutual TLS requirement.  That seems sensible to me even with this change to the `.well-known` endpoint.  The only reason you would need client authentication here is if the server were going to provide different data to different clients (including no data).  I guess you could imagine providing different endpoints to different peered providers, but in general it seems like this information should be public and uniform.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
-Events that change the state of the room are implemented through MLS proposals
-as defined in {{!RFC9420}}, thus allowing the underlying MIMI DS protocol to
-anchor the current room state cryptographically. MLS proposals are signed,
-allowing every recipient in their path to verify their authenticity.
+## Directory
 
-This document defines additional events to encapsulate MIMI DS protocol
-messages, for example, to add clients to a room's underlying MLS group or to
-request single-use key material for another user's clients.
+Like the ACME protocol {{?RFC8555}}, the MIMI protocol uses a directory document
```
bifurcation, line 582:
```suggestion
Like the ACME protocol (see {{Section 7.1.1 of ?RFC8555}}), the MIMI protocol uses a directory document
```

draft-ralston-mimi-protocol.md
bifurcation, line 621: This SHOULD should be a conditional MUST, something like:

> When the request is being made in the context of adding the target user to a room, the request MUST include ...

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
 
-The MIMI DS protocol ensures that the Hub, all follower servers and the clients
-of all participants (or at least those in the `join` state) agree on the group
-state, which includes the client list and the key material used for message
-encryption (although the latter is only available to clients). Since the group
-state also includes a copy of the room state at the time of the most recent
-commit, it is also covered by the agreement.
+The path includes the target user. The request body includes the protocol
+(currently just MLS 1.0), and the requesting user. The request SHOULD include
+the room ID for which the KeyPackage is intended, as the target may have only
```
bifurcation, line 622: Room URI?

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
+granted consent for a specific room.
 
-### Cryptographically anchoring room state {#anchoring}
+For MLS, the request includes a non-empty list of acceptable MLS ciphersuites,
+and an MLS `RequiredCapabilities` object (which contains credential types,
+non-default proposal types, and extensions) required by the requesting provider
+(these lists can be an empty). The `lastResortAllowed` field SHOULD be false.
```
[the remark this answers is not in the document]
bifurcation: I would also be fine deleting the `lastResortAllowed` flag.  We're not using it, and it's an easy thing to add back later.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
+The request body has the following form.
 
 ~~~ tls
 struct {
-   opaque user_id;
-   opaque role;
-   ParticipationState state;
-} ParticipantData
+    opaque uri<V>;
+} IdentifierUri;
```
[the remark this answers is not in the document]
bifurcation: This isn't actually hidden; it can be referenced from other sections.  TLS-PL is more of a human-readable documentation language (with unambiguous encoding semantics) than a machine-readable language.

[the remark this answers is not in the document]
bifurcation: I agree with @turt2live here that we should remain agnostic on identifiers for the moment.  To make this system hang together, we need unique IDs for rooms and users that map to their hub / home servers respectively.  We should have something detailed, but it doesn't need to be totally nailed down.

IMO what we should do here is:
* ~Flag identifiers as an open issue~ This is already done
* In this PR, use generic worridng like @turt2live suggests; don't assume that identifiers are URIs.
* In a follow-on PR, add an "Identifiers" section:
    * Define a syntax that is simple and obviously wrong:
        * Server: `example.com`
        * Room: `room-name@hub.example.com`
        * User: `user-name@home.example.com`
    * Discuss requirements and design considerations:
        * Requirements: user/room -> home/hub
        * URIs: existing syntax vs. too much complexity
        * Custom: simpler vs. have to ensure all bases covered

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
 
 struct {
-  opaque room_id;
-  ParticipantData participants<V>;
-  // TODO: Add any remaining room data
-} RoomState;
+    Protocol protocol;
+    IdentifierUri requestingUser;
+    IdentifierUri targetUser;
+    IdentifierUri roomId;
```
bifurcation, line 641: I'm probably just a type system nerd, but it might be good to distinguish `UserIdentifier` and `RoomIdentifier` types.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
+indicating failure. If the user code is success or partialSuccess, each client
+is enumerated in the response. Then for each client with a *client* success
+code, the structure includes initial keying material (a KeyPackage for MLS 1.0).
+If the client's code is `nothingCompatible`, the client's capabilities are
+optionally included (The client's capabilities could be omitted for privacy
+reasons.)
+
+If the *user* code is `noCompatibleMaterial`, the provider MAY populate the
+`clients` list. For any other user code, the provider MUST NOT populate the
+`clients` list.
```
bifurcation, line 663: I don't see `clients` defined anywhere.  Do you mean `clientCapabilities`?  Assuming that's right, the MAY / MUST NOT seems unnecessary here, since the syntax itself forbids the MUST NOT.  I would probably rephrase as "If `noCompatibleMaterial`, then the `clientCapabilities` field may be used to indicate..."

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
+
+Keying material provided from one response MUST NOT be provided in any other
```
bifurcation, line 665:
```suggestion
Key material provided from one response MUST NOT be provided in any other
```

draft-ralston-mimi-protocol.md
bifurcation: Agree with @turt2live here.  This document can just declare that two app IDs exist and what state they correspond to, and then the IANA section can assign them values.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
+happen either before or simultaneously with the corresponding MLS operation.
 
-None
+Removing an active user from a participant list or banning an active participant
+SHOULD happen simultaneously with any MLS changes made to the commit removing
+the participant.
 
-#### Send Message {#ev-sendmessage}
+A hub provider which observes that an active user has been removed or banned,
+but still has clients MUST prevent any of those clients from sending or
```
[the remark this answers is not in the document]
bifurcation: I presume "in the MLS group", but we should be clear.

I would probably be stronger than this, and declare that a proposal to removing / banning a user is invalid as long as there are any clients belonging to that user in the MLS group.  The "hub queues up proposals and requires they be implemented" model described elsewhere, this can be implemented just fine.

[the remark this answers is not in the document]
bifurcation: (Maybe you still need something like this as a backstop, for the interval in between when a removal is proposed and when it is committed.)

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
+~~~ tls
+enum {
+  reserved(0),
+  full(1),
+  compressed(2),
+  delta(3),
+  patch(4),
+  byReference(5)
+  (255)
+} RatchetTreeRepresentation;
```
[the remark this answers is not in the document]
bifurcation: I would just delete this and refer to the [representation in RFC 9420](https://rfcs.online/rfcs/rfc9420.html#ratchet-tree-extension).  AFAIK none of these are defined, and defining them would entail more work than we should do in this document (arguably should be an MLS WG project).

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
-~~~tls
 struct {
-  DSRequest send_message;
-} DSSendMessage
+  select (room.protocol) {
+    case mls10:
+      PublicMessage proposalOrCommit;
+      select (proposalOrCommit.content.content_type) {
+        case commit:
+          optional<Welcome> welcome;
```
bifurcation, line 777: Presumably this should also omit the `ratchet_tree` extension.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
 ~~~
 
-**Additional validation rules**:
-
-None
+In the first use case described in the Protocol Overview, Alice creates a Commit
+containing an AppSync proposal adding `bob@b.example`, and Add proposals for all
+Bob's MLS clients.  Alice includes the Welcome message which will be sent for
+Bob, a GroupInfo object for the hub provider, and complete `ratchet_tree`
+extension.
```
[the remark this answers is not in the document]
bifurcation: At least start with "For example, ..."

bifurcation (commented):
Sending in this review as a checkpoint; I've made it about half-way through.

Overall, I think this is the right direction, but there are several minor things to fix.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
rohan-wire: I included it to tell people not to use last resort keypackages. I'm fine deleting the field and still saying only use single-use keypackages.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
rohan-wire: That is part of the problem. We had documents with various scopes and motivations and the working group asked for a document that was clear with a limited scope. Most of our audience has not read MIMI DS. The comparison isn't necessary IMO.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
turt2live: Either works imo. More pointing out that a `SHOULD` needs an attached rationale when used.

draft-ralston-mimi-protocol.md
rohan-wire: we don't have consensus on the envelope concept you are describing, and this notion you can't change the envelope in a room. In draft-mahy-mimi-group-chat there is a single document that defines the style of conference and the permissions. I've seen at least 4 other variations. 

rather than carve that in stone, why don't we allow the hub to set the rules about what authorization is required to make certain changes. If a hub never gives the change_room_type permission, then nobody will ever be able to.

draft-ralston-mimi-protocol.md
rohan-wire: I can think of plenty of providers who would like to limit the information about their internal endpoints. Remember, only a handful of providers need to know these URLs as they are not for client use.

draft-ralston-mimi-protocol.md
rohan-wire: That's what the document already says. One applicationId is for base policy, the other for the participation list.

draft-ralston-mimi-protocol.md
turt2live: *unresolving* - this needs to be made clear in the document, sorry.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
kkohbrok: I also don't have a strong opinion on this, as long as anything that's signed (i.e. things where mTLS is not enough) includes both the endpoint and the protocol version in the TBS.

draft-ralston-mimi-protocol.md
```diff
@@ -459,6 +459,357 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+~~~
+
+~~~
+{
+  "keyMaterial":
+     "https://mimi.example.com/v1/keyMaterial/{targetUser}",
+  "update": "https://mimi.example.com/v1/update{roomId}",
+  "notify": "https://mimi.example.com/v1/notify/{roomId}",
+  "submitMessage":
+     "https://mimi.example.com/v1/submitMessage/{roomId}"
```
kkohbrok, line 600: If we do versioning via the directory, we should remove the version from the paths here.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
kkohbrok: I agree with @bifurcation here. Let's discuss usage of last resort separately.

draft-ralston-mimi-protocol.md
kkohbrok: Yes please!

draft-ralston-mimi-protocol.md
```diff
@@ -459,6 +459,357 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+optionally included (The client's capabilities could be omitted for privacy
+reasons.)
+
+If the *user* code is `noCompatibleMaterial`, the provider MAY populate the
+`clients` list. For any other user code, the provider MUST NOT populate the
+`clients` list.
+
+Keying material provided from one response MUST NOT be provided in any other
+response unless the `lastResortAllowed` field was set in the requests.
+The target provider MUST NOT provide expired keying material (ex: an MLS
```
kkohbrok, line 667: Even if we have a MUST NOT here, there should be a check on the receiving end, as sender and recipient might not agree on time.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
+enum {
+    success(0);
+    partialSuccess(1);
+    incompatibleProtocol(2);
+    noCompatibleMaterial(3);
+    userUnknown(4);
+    noConsent(5);
+    noConsentForThisRoom(6);
+    userDeleted(7);
+    (255)
```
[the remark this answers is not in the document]
kkohbrok: We shouldn't be too restrictive here. We don't want to require providers to keep lists of deleted user names around.

draft-ralston-mimi-protocol.md
```diff
@@ -459,6 +459,357 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+} KeyMaterialUserCode;
+
+enum {
+    success(0);
+    keyMaterialExhausted(1);
+    onlyLastResort(2);
```
kkohbrok, line 687: We should really leave the handling of last resort stuff for a later date.

draft-ralston-mimi-protocol.md
kkohbrok: I don't have a strong opinion here, but we I agree with @rohan-wire that we shouldn't be restrictive where we don't have to be.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
kkohbrok: I think this PR removes a section where the hub state was described as the combination of queued proposals and (commited) MLS group state. I haven't quite finished the review, but this needs to be described somewhere. Ideally at some point before this section.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
kkohbrok: Agreed. This is something we can handle at a later date.

draft-ralston-mimi-protocol.md
```diff
@@ -459,6 +459,357 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+
+~~~ tls
+enum {
+  success(0),
+  wrongEpoch(1),
+  notAllowed(2),
+  hubUnresponsive(3),
+  invalidProposal(4),
+  (255)
+} UpdateResponseCode;
```
kkohbrok, line 802: We should either discuss the semantics of each of those, or leave them out for the moment. The same goes for the other status codes below.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
-
-Each `m.room.user` event is fanned out as normal ({{fanout}}). The event MAY be
-sent to additional servers, as required by {{invites}}, {{joins}}, {{leaves}},
-{{bans}}, {{knocks}}.
+> **ISSUE:** Do we want to offer a distinction between regular application
+messages and ephemeral applications messages (for example "is typing"
+notifications), which do not need to be queued at the target provider.
 
-**Additional validation rules**:
+## Fanout Messages and Room Events
```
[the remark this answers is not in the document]
kkohbrok: For state changes it definitely needs to wait for a response for the hub, as another commit might have been faster. For application messages it's less bad, but it still alters the view of the conversation. For overall consistency, I'd suggest we don't optimize here.

draft-ralston-mimi-protocol.md
```diff
@@ -459,6 +459,357 @@ ServerA->ServerC: POST /notify/clubhouse@a.example Commit
...
+struct {
+  uint64 timestamp;
+  select (protocol) {
+    case mls10:
+      /* A PrivateMessage containing an application message,
+         a PublicMessage containing a proposal or commit,
+         or a Welcome message.                               */
+      MLSMessage message;
+      /* the hub acceptance time (in milliseconds from the UNIX epoch) */
+      uint64 timestamp;
```
kkohbrok, line 893: Do we really need two timestamps in this struct?

kkohbrok (commented):
Just a partial review for now. I'll continue over the next few days.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
-Content-Type: application/octet-stream
+To be clear, clients that are being removed SHOULD receive the corresponding
+Commit message, so they can recognize that they have been removed and clean up
+their internal state.
 
-Body
-TLS-serialized Event
+The response to a FanoutMessage contains no body. The HTTP response code
+indicates if the messages in the request were accepted (201 response code), or
+if there was an error.
```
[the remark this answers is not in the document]
bifurcation: It seems like this would simplify processing on the receive side, at least for Commit messages.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
-{{membership}}.
+The state of a room consists of its room ID, its base policy, its
+participant list (including the role and participation state of each
+participant), and the associated end-to-end protocol state (its MLS group
+state) that anchors the room state cryptographically.
 
-Additional servers MAY have the event sent to them if required by the steps
-leading up to fanout.
+While all parties involved in a room agree on the room's state during a
+specific epoch, the Hub is the arbiter that decides if a state change is valid.
```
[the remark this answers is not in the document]
bifurcation: The hub provider MUST do whatever we state as MUSTs in this specification.  But the hub is allowed to have local policy.  What else do you think the hub provider MUST do?

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
bifurcation: The citation to mimi-ds is out of place.  We should not cite that document, but instead pull any necessary mechanism in here.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
+valid previously sent proposals and apply them to the MLS group state.
 
-> **TODO**: If we plan to keep this as an independent request, it will need a
-> protocol version field.
+The MIMI DS protocol ensures that the Hub, all follower servers and the clients
+of all active participants agree on the group state, which includes the client
+list and the key material used for message encryption (although the latter is
+only available to clients). Since the group state also includes a copy of the
+room state at the time of the most recent commit, it is also covered by the
+agreement.
```
bifurcation, line 958: Similar comments here about "DS protocol".  We don't really have a DS sub-protocol any more.  Just say "MIMI".

draft-ralston-mimi-protocol.md
rohan-wire: "unresolving - this needs to be made clear in the document, sorry."

I'm not sure what you think is unclear. This document is defining a mechanism that supports the basic Alice and Bob flow with *MLS*. You have *personally* asked to remove otherwise useful features that will almost certainly be required in the MLS case based on this minimalist approach. Protocol negotiation is not even remotely required. It is **100% OUT OF SCOPE**.  Resolving.

draft-ralston-mimi-protocol.md
rohan-wire: > Do we really need two timestamps in this struct?

We need the time the hub accepts the message as discussed at the post-MIMI WG meeting lunch in Prague. What is the other timestamp you are referring to?

draft-ralston-mimi-protocol.md
turt2live: Please give me a chance to reply before resolving :/

I'm asking for the document to clarify the mechanism by which providers determine they are capable of participating in the room, and by extension their clients too. At group creation time, there needs to be a check that Alice's clients and Bob's clients are actually capable of encrypting for each other *and* that both providers (if we assume they're on separate providers) can track the appropriate state.

This is further complicated when a third user on a net-new provider is aiming to become involved. The inviter and hub server need to once again check client compatibility and provider compatibility. If the client or provider can't participate, then they should not be allowed to.

Currently, the mechanism by which this is performed is not clear in the document. It may very well be a function of MLS. If it is, text needs to be added to point to it please.

What I am *not* asking for is scope creep or for an explicit create operation in our protocol. Creation is done locally on the provider, and then invites/adds go out to future members. Those invited providers provision whatever resources they need for their clients to interact with the room, effectively creating or mirroring the room onto their systems. None of this is an operation which needs technical definition in this document.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
turt2live: (is the versioning comment on the wrong thread?)

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
turt2live: yea, +1 as recorded known issue rather than trying to solve now.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
turt2live: I'm worried that the language here gives leeway for the hub provider to enact custom and arbitrary rules. We use very few MUSTs to limit this behaviour currently.

Strengthening the overall text is likely the best direction here, rather than fixing one specific line.

draft-ralston-mimi-protocol.md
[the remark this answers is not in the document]
turt2live: what @bifurcation said :)

draft-ralston-mimi-protocol.md
turt2live: right, I'll resolve this for now pending more policy discussions with the working group. It's not something which prevents merging this PR - more of a nitpick on the language.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
-  DSRequest fetch_group_info;
-} DSFetchGroupInfo
+POST /update/{roomId}
 ~~~
 
-**Additional validation rules**:
+In MLS 1.0, any change to the room base policy document or to the participant
+list is always communicated via an `AppSync` proposal type with the appropriate
+`applicationId`. The participant list change needed to add a user MUST
+happen either before or simultaneously with the corresponding MLS operation.
```
turt2live, line 738:
```suggestion
In MLS 1.0, changes to the room's policy and participant list are done with `AppSync`
proposals. A distinct `applicationId` contains the room's policy while another contains
the participant list. When adding a user, the participant list change MUST happen 
either before or simultaneously with the corresponding MLS operation.
```

draft-ralston-mimi-protocol.md
rohan-wire: > Please give me a chance to reply before resolving :/
> 
> I'm asking for the document to clarify the mechanism by which providers determine they are capable of participating in the room, and by extension their clients too. At group creation time, there needs to be a check that Alice's clients and Bob's clients are actually capable of encrypting for each other _and_ that both providers (if we assume they're on separate providers) can track the appropriate state.

If they implement what is in this PR merged into this document there is no issue. Group creation is out of scope. MIMI specs only cover MLS. In MLS, Alice can see via Bob's KeyPackage if Bob supports the necessary features.

As for adding future negotiation for other encryption or MIMI extensions, for MLS the KeyPackages can contain extensions; support can be statically configured (quite appropriate for a non-standard encryption binding); and the discovery mechanism can be extended. And no, we do not need to say that in the document.

> This is further complicated when a third user on a net-new provider is aiming to become involved. The inviter and hub server need to once again check client compatibility and provider compatibility. If the client or provider can't participate, then they should not be allowed to.
>
> Currently, the mechanism by which this is performed is not clear in the document. It may very well be a function of MLS. If it is, text needs to be added to point to it please.

Then once this PR is merged, I would be delighted to see issues and PRs which point out specific use cases or problemss. Until then I would advise you to carefully read the sections in RFC 9420 (section 13 - especially 13.4) and draft-ietf-mls-architecture (sections 6.8, 7, and 8.2.4) on MLS extensibility.
 
> What I am _not_ asking for is scope creep or for an explicit create operation in our protocol. Creation is done locally on the provider, and then invites/adds go out to future members. Those invited providers provision whatever resources they need for their clients to interact with the room, effectively creating or mirroring the room onto their systems. None of this is an operation which needs technical definition in this document.

Respectfully, you are asking for a generic extensibility negotiation mechanism, and/or a lot of text that does not fit the current scope of this PR and the target draft. That is not appropriate. It is out of scope.

turt2live: > I would advise you to carefully read the sections in RFC 9420 (section 13 - especially 13.4) and draft-ietf-mls-architecture (sections 6.8, 7, and 8.2.4) on MLS extensibility.

Please add these references to the document if they address the concern. It is not clear as a reader that these sections are required reading.

I do not agree that people reading this document/PR will inherently know that the mechanism for ensuring compatibility is in MLS.

turt2live: To reiterate though: I'm really just looking for a sentence or two *somewhere* to show we've thought about this, because clearly we have. The scope of this PR is not clear to me, but from the title I would consider Alice fetching Bob's KeyPackage and deciding that communication is impossible as a 'basic flow'.

turt2live: If we think it's out of scope for this PR, please consider https://github.com/bifurcation/ietf-mimi-protocol/pull/51

draft-ralston-mimi-protocol.md
kkohbrok: That's fair, but if the `v1` in the path is not relevant, we might want to choose a less confusing example.

draft-ralston-mimi-protocol.md
kkohbrok: The first field of this struct is labeled `timestamp` and then there's another one on the line this comment is on. I do agree that we generally need a timestamp. I was just confused by the presence of two timestamps.

draft-ralston-mimi-protocol.md
rohan-wire: pretended we already had two specific applicationIds: `mimiRoomPolicy` and `mimiParticipantList`

draft-ralston-mimi-protocol.md
rohan-wire: issue #53

draft-ralston-mimi-protocol.md
rohan-wire: search for this vector. it's there
`ClientKeyMaterial clients<V>;`

draft-ralston-mimi-protocol.md
rohan-wire: There are tons of uses of both key material and keying material. we can decide later which of these we want to use. for now i will ignore this.

draft-ralston-mimi-protocol.md
```diff
@@ -560,894 +560,419 @@ To put these changes into effect, a client or server encodes them in an AppSync
...
-
-4. If the event is invalid, it is rejected. Otherwise, the event is fanned out
-   ({{fanout}}) to all participating servers, plus the joiner's server as they
-   are now participating in the room too.
 
-If the user was added to the room via a standalone `m.room.user` event, the
-user's clients are able to add themselves to the cryptographic group state via
-one or more `ds.commit` events after fetching the room's current information via
-a `ds.fetch_group_info` event.
+If the protocol is MLS 1.0, the request body is an MLSMessage with a WireFormat
```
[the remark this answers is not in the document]
rohan-wire: #52

draft-ralston-mimi-protocol.md
rohan-wire: discussed in #52

kkohbrok (approved):
Thanks for making the changes @rohan-wire! I'd say let's get this merged and then we can see what issue we want to follow up with.

PR #51, opened 2024-02-14
Clarify that KeyPackages *are* the negotiation mechanism for support

turt2live:
ref https://github.com/bifurcation/ietf-mimi-protocol/pull/50/files#r1486736385

### DIFF
--- a/draft-ralston-mimi-protocol.md
+++ b/draft-ralston-mimi-protocol.md
@@ -545,6 +545,11 @@ KeyPackageRef values for the returned KeyPackages, and the identity of the
 provider from which they were received.  This information is then used to route
 Welcome message to the proper provider.
 
+The returned KeyPackages identify whether the device (and therefore user) is
+capable of participating in the group. If compatible, normal MLS operations are
+used to add the device alongside a participation list change proposal for the
+user (described below).
+
 ### Participation List Changes
 
 The participation list can be changed by adding or removing users.  These


rohan-wire:
I like the spirit of this PR, but it is not technically correct. The requester asks ahead of time for KeyPackages with whatever the required extensions, proposals, and credentials are. 

From PR #50, note the `RequiredCapabilities` line from the struct below and the text about what it is for.
```
struct {
    Protocol protocol;
    IdentifierUri requestingUser;
    IdentifierUri targetUser;
    IdentifierUri roomId;
    select (protocol) {
        case mls10:
            CipherSuite acceptableCiphersuites<V>;
            RequiredCapabilities requiredCapabilities;
            bool lastResortAllowed;
    };
} KeyMaterialRequest;
```

"For MLS, the request includes a non-empty list of acceptable MLS ciphersuites, and an MLS RequiredCapabilities object (which contains credential types, non-default proposal types, and extensions) required by the requesting provider (these lists can be an empty)."

bifurcation:
DT call 2024-02-28: 
* @rohanmahy's concern is expressing the `required_capabilities` part
* @bifurcation suggests this is good in the key material fetch section
* @turt2live will look on Friday, @rohanmahy may look at it sooner

turt2live:
Closing in favour of https://github.com/bifurcation/ietf-mimi-protocol/pull/63

PR #62, opened 2024-03-01
Introduce very lightweight notion of role-based access control

rohanmahy:


turt2live (approved):

draft-ralston-mimi-protocol.md
```diff
@@ -217,11 +218,16 @@ This operation is local to the service provider, and does not entail any MIMI
...
+For authorization purposes, MIMI uses permissions based on room-defined roles.
+For example, a room might have a role named "admin", which has `canAddUser`,
+`canRemoveUser`, and `canSetUserRole` permisions.
+
 Here, we assume that Alice uses ClientA1 to create a room with the following
 properties:
 
 * Room Identifier: `mimi://a.example/r/clubhouse`
-* Participants: `[mimi://a.example/u/alice]`
+* Roles: `["admin", [canAddUser, canRemoveUser, canSetUserRole]]`
```
bifurcation, line 229: Suggest enclosing the permissions in double quotes as well.  Or removing them from the label name.  Maybe something like:

```
admin = [canAddUser, canRemoveUser, canSetUserRole]
```

draft-ralston-mimi-protocol.md
```diff
@@ -562,11 +568,13 @@ Welcome message to the proper provider.
...
 
 The participation list can be changed by adding or removing users.  These
-changes are described without a specific syntax as a list of adds and removes:
+changes are described without a specific syntax as a list of adds, removes,
+and updates:
 
 ~~~ ascii-art
 Add: ["mimi://d.example/u/diana", "mimi://e.example/u/eric"],
 Remove: ["mimi://b.example/u/bob"],
+Update: [["mimi://c.example/u/cathy", "admin"]]
```
bifurcation, line 577: What are the semantics of Add with respect to roles?  I assume every user has to have an assigned role.  Are new users assigned a default role, or does the Add need to specify a role?

Update has an unfortunate resonance with the MLS operation of the same name (and different semantic).  I would suggest using `SetRole` for this, which aligns well with the example permissions above.

draft-ralston-mimi-protocol.md
```diff
@@ -562,11 +568,13 @@ Welcome message to the proper provider.
 ### Participation List Changes
 
 The participation list can be changed by adding or removing users.  These
-changes are described without a specific syntax as a list of adds and removes:
+changes are described without a specific syntax as a list of adds, removes,
+and updates:
```
bifurcation, line 572: You need to elaborate a little more here and above.

* The base policy is probably where the role definitions go
* The participation list has a list of participants and their roles (cf. "permissions" now)
* The participation list can be modified by adding users, removing users, **or setting a user's role**

Also, "participant list" seems more natural to me than "participation list"

bifurcation (approved):

draft-ralston-mimi-protocol.md
rohanmahy: OK updated.

draft-ralston-mimi-protocol.md
rohanmahy: ok modified as suggested

PR #72, opened 2024-03-13
Add consent discussion and mechanism

rohanmahy:


draft-ralston-mimi-protocol.md
```diff
@@ -1246,6 +1245,98 @@ key. The GroupInfo in another context might be sufficiently sensitive that
...
+implicitly determine consent. This section describes a mechanism by which providers can explicitly request consent from a user of another provider,
+cancel such a request, convey that consent was granted, or convey that
+consent was revoked or preemptively denied.
+
+~~~
+POST /requestConsent/{targetDomain}
+POST /updateConsent/{requesterDomain}
+~~~
+
+A `requestContent` request is used by one provider to request explicit
```
kkohbrok, line 1260:
```suggestion
A `requestConsent` request is used by one provider to request explicit
```

draft-ralston-mimi-protocol.md
```diff
@@ -1246,6 +1245,98 @@ key. The GroupInfo in another context might be sufficiently sensitive that
...
+} ConsentEntry;
+
+struct {
+  IdentifierUri requesterUri;
+  IdentifierUri targetUri;
+  optional<RoomId> roomId;
+} ConsentScope;
+~~~
+
+An `updateContent` request is used by one provider to provide explicit
```
kkohbrok, line 1307:
```suggestion
An `updateConsent` request is used by one provider to provide explicit
```

kkohbrok (commented):
Thanks for writing this up! Sounds generally like a good mechanism to me. Two thoughts:

Getting hold of a KeyPackage is generally pretty easy (e.g. just pluck it from a tree in some group you're in), so it might be worth restricting their use a bit. As I understand this proposal, it's mostly about the decision whether the a provider will hand out KeyPackages for its users. However, we could extend it by generally requiring that when adding a new user to a room (with one or more initial clients), the adder has to prove that the target user (or at least the target user's provider) has consented, in which case the scoping with the room id becomes enforcable.

I'm also wondering if we can avoid leaking the identity of the KeyPackage requester to the target's provider, i.e. prove that consent was previously obtained in zero knowledge of the requester's identity. We could use anonymous credentials or something like Privacy Pass to that end.

If either of those suggestions sounds reasonable I'd be happy to contribute either to this PR or in a follow-up.

rohanmahy:
Hi Konrad,
Thanks for the fast review.

Regarding the first suggestion, I always wanted providers to be able
restrict KeyPackage usage to a specific room, but I didn't want to make it
mandatory, since some providers have other implicit consent mechanisms. I'm
happy to discuss with you and see how we can improve the text.

I would love to see a PR improving the privacy as you described.I think
somewhere it says that at least that the user identiifiers can be
pseudonymous user IDs only used for a single room.

Thanks,
-rohan

kkohbrok:
There's not need to make the scoping mandatory. However, it would be a bit more involved than just changing the text a little bit. Here's what I roughly had in mind:

- Whenever a provider responds with a KeyPackage, it also provides as proof of consent a signature over the KeyPackage (or hash thereof), as well as a over the scope of the consent and the validity period of the proof of consent. The scope might include the name of the requester (or a pseudonym or similar) and optionally a room id.
- If the consent mechanism is implicit, the provider can just sign the KeyPackage without additional information
- Whenever a KeyPackage is used to add someone to a group, the adder has to provide the signature and scope of consent. - Other group members can then verify that the adder is actually allowed to use the KeyPackage in this context by verifying that the senders name/pseudonym, the room id and the validity period.

We could also use this mechanism to ensure that a requester always adds all clients of a given user: When a user requests KeyPackages for a user, the provider would provide a signature over all of the KeyPackages in the response.

I'll try to come up with a proposal for how we can hide requester Metadata in a follow-up PR.

draft-ralston-mimi-protocol.md
```diff
@@ -1246,6 +1245,98 @@ key. The GroupInfo in another context might be sufficiently sensitive that
...
+
+The response to an `updateConsent` is usually a 201 Accepted (indicating
+the `updateConsent` was received), optionally a 404 Not Found (indicating the
+`requesterUri` is unknown), or a 500-class response. The response code
+merely indicates that the request was received. A provider that does not
+wish to reveal if a user is not found can respond with a 201 Accepted.
+
+> **NOTE**: Blocking a user might be privacy sensitive. If this is the case
+> the target provider does not need to send a `revoke` to inform the
+> requester provider.
```
turt2live, line 1338: Block means neither user can communicate with the other, and both users are aware of that fact. To prevent a user from communicating with you, but them not being aware of that fact, a "mute" is used instead. These terms are defined by the UK Online Safety Act and similar proposed legislation in the EU, US, and Canada.

Or in other words, privacy sensitivity is covered by mutes, not blocks.

draft-ralston-mimi-protocol.md
```diff
@@ -1246,6 +1245,98 @@ key. The GroupInfo in another context might be sufficiently sensitive that
...
+struct {
+  IdentifierUri requesterUri;
+  IdentifierUri targetUri;
+  optional<RoomId> roomId;
+} ConsentScope;
+~~~
+
+An `updateContent` request is used by one provider to provide explicit
+notice from a target user at one provider that consent for a specific
+"requester" was granted, revoked, or preemptively denied. In this context,
```
turt2live, line 1309: It feels a bit awkward that a server needs to round trip another request to decline consent, if it already knows at the point of request that it will be declined. For example, the target user has muted/blocked the requesting user or the server has a safety policy which denies consent requests from the requesting server. Could we add an error code/indication that consent is not possible to `requestConsent` to short circuit the need for `requestConsent->201, updateConsent`?

draft-ralston-mimi-protocol.md
```diff
@@ -1246,6 +1245,98 @@ key. The GroupInfo in another context might be sufficiently sensitive that
...
+  request(1),
+  grant(2),
+  revoke(3),
+  (255)
+} ConsentOperation;
+
+struct {
+  ConsentOperation consentOperation;
+  IdentifierUri requesterUri;
+  IdentifierUri targetUri;
```
turt2live, line 1292: These fields likely need a condition to say they belong to the respective server. Namely, the requester belongs to the requesting server and the target belongs to the receiving server. Otherwise, 400 Bad Request for making an improper request.

draft-ralston-mimi-protocol.md
```diff
@@ -1246,6 +1245,98 @@ key. The GroupInfo in another context might be sufficiently sensitive that
 it should be encrypted from the end client to the hub provider (unreadable
 by the local provider).
 
+## Convey explicit consent
```
turt2live, line 1248: Do these endpoints get routed through the hub?

turt2live (commented):
overall this looks like a great starting point for a larger conversation. I'm still pondering the safety implications of such a system, but the basics do seem to be covered.

draft-ralston-mimi-protocol.md
rohanmahy: Thanks for asking. I think they do not. Alice's domain would need to have a consent relationship of some type directly with the domain of the target client.

The KeyPackage request gets routed through the hub, so it can route the subsequent Welcome.

draft-ralston-mimi-protocol.md
rohanmahy: OK, I think I can say that in prose. How about?

"For a `requestContent`, the `targetUri` needs to be in one of the domains of the receiving server, and the `requesterUri` needs to be in one of the domains of the sending server. 

For `updateConsent` the `requesterUri` needs to be in one of the domains of the receiving server, and the `targetUri` needs to be in one of the domains of the sending server."

draft-ralston-mimi-protocol.md
rohanmahy: Thanks for pointing that out. I think there is a lot of usage of these terms predating the [2023 UK Online Safety Act](https://www.legislation.gov.uk/ukpga/2023/50/enacted). Block is used by XMPP and WhatsApp, and I think mute could be confused for audio/video muting, muting notifications, or revoking voice in a group chat. I don't have a strong feeling about which set terms we use.

I'll make a slide.

draft-ralston-mimi-protocol.md
rohanmahy: I think the may several of these systems work, the requester never finds out if consent was rejected or the user just never responded (for privacy reasons). I think it is fine to say that a server can just come right out and say no, you don't get consent if this is consistent with their privacy policies. However the most likely case seems that they never get an answer.

I expect in some systems updateConsent with a reject will only be used to remove consent after it had been previously granted.

turt2live: hmm, good point. Most requests would be ignored - given the rarity, two requests seems fine.

Thanks for working through the suggestion. Resolving.

draft-ralston-mimi-protocol.md
turt2live: Looks good to me, though the plural of "domains" is a bit scary - not sure we've discussed a provider having multiple domains yet.

draft-ralston-mimi-protocol.md
turt2live: (this makes sense - it should be clarified in the text)

draft-ralston-mimi-protocol.md
rohanmahy: OK, changed "Blocking" to "Revoking consent".

draft-ralston-mimi-protocol.md
rohanmahy: OK, add a short paragraph in https://github.com/bifurcation/ietf-mimi-protocol/pull/72/commits/dd9199ea036c66081e1a251cbf17746efeb7860e . Hopefully that does the trick.

bifurcation:
Interim 2024-04-10:
* Clear to merge once conflicts are resolved.

rohanmahy:
Replaced by #78 . Closing.

PR #73, opened 2024-03-13
Add basic abuse reporting mechanism

rohanmahy:


draft-ralston-mimi-protocol.md
```diff
@@ -1246,6 +1246,59 @@ key. The GroupInfo in another context might be sufficiently sensitive that
 it should be encrypted from the end client to the hub provider (unreadable
 by the local provider).
 
+## Report abuse
+
+Abuse reports are only sent to the hub provider. They are sent as an HTTP
+POST request.
```
turt2live, line 1252: Feels a bit weird that abuse would only be reported to the hub. The provider which sent the content should receive that report, and there may need to be redirection from the server to the moderators of the room.

There's also a whole can of worms opened here regarding safety legislation, including reversible actions, reporting, appeals, suspension, and notification of action taken (both to the reporter and affected user).

rohanmahy: >Feels a bit weird that abuse would only be reported to the hub. 
What does the MIMI protocol need to provide at minimum?  Let's consider the reporting client, the reporting provider, the hub, the abuser's provider, and the abuser's client.

The clients don't run the MIMI protocol. The reporting provider got informed using a provider-specific protocol, the hub (especially in it's role of policy enforcer) definitely needs to get this information. So that leaves the abuser's provider.

>The provider which sent the content should receive that report, and there may need to be redirection from the server to the moderators of the room.

I will argue that the abuser provider shouldn't automatically get the report, and in some cases should never get the (unredacted) report. The peering relationship between the hub and the abuser provider should specify the details of what the abuser's provider gets. Also, after the report is sent, the hub may do all kinds of analysis and/or send the content to a human moderator to look at. Only after the hub provider decides something is or is not in their policy would they have any cause to send something to another provider. In any case, I think this need not be part of a real-time protocol in MIMI. 

I think all the actions you listed in your can of worms refer to a subset of the information in the original report.

alvestrand:
In the case where the abuse comes from the abuser's provider (such as when the "abuser" is a fake identity created by the abuser's provider in order to achieve abusive outcomes like spam or phishing), sending an abuse complaint verbatim to the abuser's provider is exactly the wrong thing to do.

The abuser's provider may need to get told that sanctions have been imposed against his user, but telling him why feeds into machine learning on how to avoid anti-abuse mechanisms.

draft-ralston-mimi-protocol.md
turt2live: (the quotes look a bit mangled, ftr)

I think we're jumping to an edge case here. Reports over federation *should not* contain details of the reporter, for several safety reasons. Following the path of a report:

* Alice reports Bob's message to their local provider. The local provider receives information about Alice's client, Bob's message, decryption keys, and possibly a user-supplied label/reason for the report, among other details. As this is a provider-local operation, it is out of scope for MIMI.
* Alice's local provider does some filtering and processing on the report. This may be automated, or it may be subject to human review. In this case, the report cannot be handled by Alice's local provider, and so Bob's provider needs to be involved.
* Alice's local provider creates a *minimal* report to send to Bob's server. Specific details include the message identifier which was reported, maybe decryption keys (assuming they don't contain identifying information of the reporter), and optionally a classification label/reason. The label/reason *may* be different from the original report.
* Bob's local provider upon receiving this report processes it, potentially maliciously forwarding it to Bob directly. This is why we don't include information about who did the report, just where it came from. 

I've not indicated which server is the hub here because I don't think it applies. Forwarding all reports to the hub feels like a risk, as the hub provider is then responsible for all abuse mitigation for the room - not something MIMI's consumers are likely to agree with. Instead, reports can be sent directly to the server which emitted the content, removing the need for a platform moderation team to monitor reports on platforms they don't control.

rohanmahy:
Discussion about the potential for a reporter to forge an abusive message at interim meeting 10-Apr-2024:

- Travis says the abuse reports should go to the reporter's provider, while the hub provider is specified in the PR currently. Harald says you shouldn't reveal abuse reports to the alleged abuser's provider.
- The situation is better than if we only have AEAD, but we still may need some additional franking (by the hub using a symmetric algorithm)
- There is a risk of "outing" a pseudonymized user.

rohanmahy:
Replacing with franking based proposal in #83

PR #74, opened 2024-03-13
Initial stab at finding internal identifiers primitive

rohanmahy:


draft-ralston-mimi-protocol.md
turt2live, line 1: This feature feels like it should be part of the discovery document rather than protocol. It feels distinct enough to be a dedicated document with search capabilities and global directories.

rohanmahy: The discovery documents are explicitly about discovering _which provider_ is associated with a specific SII. This is entirely about searching inside a provider, instead of finding a provider, so I think it is still in scope. This is a good question for the WG though.

draft-ralston-mimi-protocol.md
```diff
@@ -1246,6 +1246,87 @@ key. The GroupInfo in another context might be sufficiently sensitive that
...
 by the local provider).
 
+## Find internal address
+
+The identifier query is to find the internal URI for a specific user on a
+specific provider. It is only sent from the local provider to the target
+provider (it does not transit a hub).
+
+~~~
+GET /identifierQuery/{domain}
```
tgeoghegan, line 1256: GET is inappropriate since there is a request body. I think POST or the not-yet-quite-standard QUERY methods would work.

rohanmahy: The QUERY method is described in https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/
The draft has expired twice since becoming a WG item and three times as an individual draft. As of today it has 88 [open issues](https://github.com/httpwg/http-extensions/issues) in the github repo. Even though I think this usage is 100% aligned with QUERY, I don't want to depend on this work which appears to be proceeding slowly.

I added a commit to just use POST instead of GET.

tgeoghegan: I think that's a sensible choice. You might add a note that this POST request is idempotent and safe in the sense defined by [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#idempotent.methods).

draft-ralston-mimi-protocol.md
```diff
@@ -1246,6 +1246,87 @@ key. The GroupInfo in another context might be sufficiently sensitive that
 it should be encrypted from the end client to the hub provider (unreadable
 by the local provider).
 
+## Find internal address
+
+The identifier query is to find the internal URI for a specific user on a
+specific provider. It is only sent from the local provider to the target
+provider (it does not transit a hub).
```
rohanmahy, line 1253:
```suggestion
provider (it does not transit a hub). Note that this POST request is idempotent and safe in the sense defined by {{Section 9.2.2 of RFC9110}}.
```

draft-ralston-mimi-protocol.md
```diff
@@ -1246,6 +1246,87 @@ key. The GroupInfo in another context might be sufficiently sensitive that
 it should be encrypted from the end client to the hub provider (unreadable
 by the local provider).
 
+## Find internal address
+
+The identifier query is to find the internal URI for a specific user on a
+specific provider. It is only sent from the local provider to the target
+provider (it does not transit a hub). Note that this POST request is idempotent and safe in the sense defined by {{Section 9.2.2 of RFC9110}}.
```
tgeoghegan, line 1253: nit: wrap the long line

tgeoghegan (approved):

PR #75, opened 2024-03-13
Add a download files proxy primitive

rohanmahy:


draft-ralston-mimi-protocol.md
```diff
@@ -1246,6 +1246,23 @@ key. The GroupInfo in another context might be sufficiently sensitive that
 it should be encrypted from the end client to the hub provider (unreadable
 by the local provider).
 
+## Download files
```
turt2live, line 1249: this is a WIP PR, but some design considerations from prior learnings:

* Files should be linked to a specific message or event, and access controls implemented over the proxyDownload request. Specifically, the requesting server MUST have visibility on the message in order to have the download succeed. This is to prevent the remote server from becoming an open proxy and effective CDN (typically for illegal content).
* The request must include sufficient authorization information to complete the ACL check above. This doesn't require identifying the specific user/client, but must be able to identify the server. (I think we already support this, but noting it anyways just in case).
* The hub server, who is presumably receiving this request, should make use of a [leaky bucket algorithm](https://en.wikipedia.org/wiki/Leaky_bucket#As_a_meter) to minimize bandwidth costs.

rohanmahy:
Once we have more room policy established, we can have attachments policies. One of these could specify if attachments are to be stored in the local sender's provider domain or the hub's domain. The download proxy primitive could then be restricted to only working for URLs in the specified domain, or even restricted to a specific path.

rohanmahy:
Closing in favor of #111

PR #79, opened 2024-06-10
Semi-private handshake messages in Update Room

rohanmahy:
According to room policy, members could send handshake messages (Commits and Proposals) as PublicMessage or PrivateMessage. When a PrivateMessage the key and nonce used to encrypt the PrivateMessage is HPKE encrypted for the hub and included in the message.
the HPKE `Seal<Base>` `info` is the hash of the complete MLSMessage that would have been sent.


rohanmahy:
replaced by PR #80

PR #80, opened 2024-06-13
Semi-private handshake can use an MLS SemiPrivateMessage 

rohanmahy:
WIP

Alternate (to #79 ) way to have a semi-private handshake (private, plus shared encrypted with the hub).


rohanmahy:
Per the discussion at IETF120 and the subsequent consensus call, the group will move forward with implementation of SemiPrivateMessage as a baseline mechanism and pseudonyms as an opt-in feature (which should require little or no new mechanism for the hub and other members/participants).

This PR replaces #79

PR #83, opened 2024-07-21
Add abuse reporting mechanism with message franking

rohanmahy:


draft-ietf-mimi-protocol.md
```diff
@@ -53,6 +53,44 @@ author:
 normative:
 
 informative:
+  SecretConversations:
+    target: https://about.fb.com/wp-content/uploads/2016/07/messenger-secret-conversations-technical-whitepaper.pdf)
```
turt2live, line 57:
```suggestion
    target: https://about.fb.com/wp-content/uploads/2016/07/messenger-secret-conversations-technical-whitepaper.pdf
```

draft-ietf-mimi-protocol.md
```diff
@@ -1461,6 +1625,52 @@ struct {
 > {{?I-D.mahy-mimi-identity}}. Any specific conventions which are needed
 > should be merged into this document.
 
+## Report abuse
+
+Abuse reports are only sent to the hub provider. They are sent as an HTTP
```
turt2live, line 1630: What is the intended scope of this reporting mechanism? Sending things to the hub feels improper when trying to report platform-level concerns ("Alice has a pattern of disruptive behaviour"), but would be fine when aiming to report messages without a broad pattern of abuse. 

We may need two (or more) mechanisms to cover the use cases. Each mechanism should describe which use case(s) they cover specifically.

turt2live (approved):
The idea looks good, though I can't speak to the cryptography/franking specifics. +1 to franking for sure.

I also have the ongoing (and outstanding) concern about what use cases we're applying to the abuse reporting mechanisms.

draft-ietf-mimi-protocol.md
turt2live: From IETF120: this mechanism is specifically to allow the hub to enforce policy at the room/group level. Broad patterns of abuse ("provider A's terms of service are being violated") would be handled elsewhere. 

Examples of policy would be "no cursing", where the hub can then handle the report as needed.

Clarifying this in the PR is appreciated.

rohanmahy:
I just merged a change which makes the franking_context safe from metadata inspection by follower providers. It uses a secret derived using the "Associated Parties key schedule" documented in draft-kohbrok-mls-associated-parties. I think this closes the last open issue with the franking mechanism.

draft-ietf-mimi-protocol.md
```diff
@@ -1186,10 +1200,12 @@ follower providers.
 When a client receives and decrypts an otherwise valid application message
 from a hub provider, the client looks for the existence of a frank
 (consisting of the `franking_tag` in the AAD, the `serverFrank` and the
-`franking_context_hash`. If so it verifies the construction of the
-`franking_tag` from the content of the message, and the construction of
-the `franking_context_hash` from the sender URI, room ID, and
-`acceptedTimestamp`.
+`franking_context_hash`. If so, it derives the `franking_context_secret`
+from the `ap_exporter_secret` in the Associated Party Key Schedule
+{{I-D.kohbrok-mls-associated-parties}}; then it verifies the construction of
```
dconnolly, line 1205: How is this done, exactly? `draft-kohbrok-mls-associated-parties` has no mention of `franking_context_secret` or non-explicit derivations in its key schedule

rohanmahy: In Figure 1 of Associated Parties, the ap_exporter_secret is derived from the ap_epoch_secret for the Hub. `franking_context_secret` is derived from `ap_exporter_secret` using the label "franking_context" (just added the label). See lines 1174 to 1178 for the discussion of the derivation.

kkohbrok (approved):

draft-ietf-mimi-protocol.md
dconnolly: OK; I may submit a PR to this one just to clarify the derivation even more; I wonder if Associated Parties can/should provide explicit guidance on how to do this sort of thing 'in general'

draft-ietf-mimi-protocol.md
```diff
@@ -993,14 +1057,21 @@ struct {
     case mls10:
       /* PrivateMessage containing an application message */
       MLSMessage appMessage;
+      IdentifierURI sendingUri;
```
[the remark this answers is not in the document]
rohanmahy: I'll do a massive case realignment PR either before or after WGLC

PR #86, opened 2024-10-21
Encrypt GroupInfo and ratchet_tree in repsonse to /groupInfo endpoint

rohanmahy:
… for better metadata privacy

### DIFF
--- a/draft-ietf-mimi-protocol.md
+++ b/draft-ietf-mimi-protocol.md
@@ -1184,6 +1184,7 @@ struct {
       CipherSuite cipher_suite;
       SignaturePublicKey requestingSignatureKey;
       Credential requestingCredential;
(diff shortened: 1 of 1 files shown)


kkohbrok (approved):

draft-ietf-mimi-protocol.md
```diff
@@ -1184,6 +1184,7 @@ struct {
       CipherSuite cipher_suite;
       SignaturePublicKey requestingSignatureKey;
       Credential requestingCredential;
+      HPKEPublicKey groupInfoPublicKey;
```
bifurcation, line 1187: At this point, you could just send a KeyPackage and have the response be a Welcome with either a Welcome extension or a PrivateMessage.

Alternatively, you could just use OHAI.

bifurcation (approved):

PR #87, opened 2024-10-21
Diagram the extended Associated Parties Key Schedule to derive franking_context_secret

dconnolly:
@rohan-wire @kkohbrok


rohanmahy (approved):

dconnolly:
@rohanmahy I think the workflow run needs to be approved and then this can be merged into the message franking branch `abuse2`

Issue #91, opened 2024-11-17
Move FrankingAssertion map to mimi-content


rohanmahy:
The FrankingAssertion map in Section 5.4.1.1 Client creation and sending should be defined in draft-ietf-mimi-content

rohanmahy:
Need comment about relationship between the two documents @rohanmahy

rohanmahy:
This is now in mimi-content PR46.

PR #94, opened 2024-12-16
Allow searches for multiple fields

rohanmahy:
- Allow searches for multiple fields (with AND semantics)
- Clarify the semantics of SearchIdentifierType enum

Partially addresses Issue #91

### DIFF
--- a/draft-ietf-mimi-protocol.md
+++ b/draft-ietf-mimi-protocol.md
@@ -1723,15 +1723,13 @@ format of a telephone number with the "+" prefix matches exactly (for example:
 of any field which contains the name of a (usually human) user. For example,
 `mat` would match first (given) or middle names Matt, Matthew, Mathias, or
 Mathieu and last (family) names of Mather and Matali. `wholeProfile` means that
-the query string matches a case-insensitive substring of any searchable field in
+the query string matches a substring of any searchable field in
 a user's profile.
 
 `oidcStdClaim` means that the query string exactly matches the specified
 UserInfo Standard Claim (defined in Section 5.1 of {{OidcCore}}).
 `vcardField` means that the query string exactly matches the specified vCard
 property listed in the vCard Properties IANA registry.
-
-> **TODO**: Add language about case sensitivity for some other identifier types.
 
 As noted above, searches only return results for a user when the fields searched
 are searchable according the user's and provider's search policies.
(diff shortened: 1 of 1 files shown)


draft-ietf-mimi-protocol.md
```diff
@@ -1687,11 +1701,41 @@ struct {
     case oidcStdClaim:
       opaque claimName<V>;
     case vcardField:
-      opaque fieldName<V>;
+      opaque propertyName<V>;
```
bifurcation, line 1704: Nit: I would put this `select` above the `searchValue`, since the colloquial order is `(key, value)`.

draft-ietf-mimi-protocol.md
```diff
@@ -1687,11 +1701,41 @@ struct {
...
   };
+} QueryElement;
+
+struct {
+  QueryElement query_elements<V>;
```
bifurcation, line 1709: What is the semantic of having multiple query elements?  Are you asking for a profile that matches all of them?  Any of them?

draft-ietf-mimi-protocol.md
```diff
@@ -1687,11 +1701,41 @@ struct {
...
+exactly, for example (`asmith@a.example`). `phone` means the international
+format of a telephone number with the "+" prefix matches exactly (for example:
+`+12125551212`).
+
+`partialName` means that the query string matches a case-insensitive substring
+of any field which contains the name of a (usually human) user. For example,
+`mat` would match first (given) or middle names Matt, Matthew, Mathias, or
+Mathieu and last (family) names of Mather and Matali. `wholeProfile` means that
+the query string matches a case-insensitive substring of any searchable field in
+a user's profile.
```
bifurcation, line 1727: Case-insensitivity us normally never a good idea, but here I think you might not be loose enough.  For example, if the query contains `Mathtew`, is the responding provider allowed to spell-check that to `Matthew`?  I would probably be inclined to be very loose about the request semantics, and instead have the response provide the attributes that the responding provider thought matched, so that then the requesting provider / user can decide if the match is valid.

draft-ietf-mimi-protocol.md
rohanmahy: All of them. there was text somewhere but I can restate here

draft-ietf-mimi-protocol.md
rohanmahy: fair enough

PR #101, opened 2025-01-13
Add section on minimal metadata rooms

kkohbrok:
This PR introduces the notion of pseudonymous credentials for MIMI by introducing a new room type "Minimal Metadata Rooms". We could also make that a configuration flag and/or use another name. I'm open to suggestions in that regard. There are still a few TODOs/open questions, though:

- Encryption of `identity_link_keys`: We intend to solve this via the TreeWrap mechanism we proposed, but we haven't quite finished writing up the spec for that. For now it's a simple static key.
- External joins: There's still a small issue with external joins. In theory, new joiners just need the `identity_link_wrapper_key` to join the group, but if we start rotating that, we run into race conditions. How we solve that depends a bit on the situations in which we need external joins. Maybe we can discuss this in a future interim.
- Connections: For Minimal Metadata Rooms to work, users need to learn the `connection_key` of other users they want to add to a group. Since we probably also want a notion of "connection" for other things such as consent, this PR only expresses the requirement for connection key exchange for now.

### DIFF
--- a/draft-ietf-mimi-protocol.md
+++ b/draft-ietf-mimi-protocol.md
@@ -1754,6 +1754,192 @@ struct {
 
 There is no response body. The response code only indicates if the abuse report was accepted, not if any specific automated or human action was taken.
 
+# Minimal metadata rooms
+
+The room state is visible to the hub and with it the room's participant list,
+giving the hub access to a significant amount of user metadata.
+
+To limit the amount of metadata the hub has access to, rooms can be created as
+_minimal metadata rooms_ (MMR). In an MMR the participant list and the
+credentials in the room's underlying MLS group consist only of pseudonyms. The
+real identifiers are stored alongside the pseudonyms encrypted under a key known
+only to room participants, but not the hub.
+
+MMRs requires some additional key management, which leads to restrictions in how
+the MMR can be joined and which users each participant can add to the room.
+
+## Credential encryption
+
+Identifiers of participants and their clients occur in two locations in a room's
+state: the participant list and the credentials of the room's underlying MLS
+group. In an MMR, the real identifiers of clients and users are replaced by
+pseudonyms in the shape of random UUIDs qualified with the domain of the user's
+provider.
+
+In MMRs, all leaves of the underlying group MUST contain PseudonymousCredentials.
+
+~~~ tls
+struct {
+  IdentifierUri client_pseudonym;
+  IdentifierUri user_pseudonym;
+  opaque signature_public_key;
+  opaque identity_link_ciphertext<V>;
+} PseudonymousCredential
+~~~
+
+- `user_pseudonym`: The pseudonym of the client's user in this group
+- `client_pseudonym`: The pseudonym of the client identified by this credential
+- `signature_public_key`: The signature public key used to authenticate MLS
+  messages
+- `identity_link_ciphertext`: A ciphertext containing a credential with the
+  clients real identifier
+
+
+In each room, the `user_pseudonym` of a client MUST be the same across all
+clients of a user and it MUST be the same as the user's entry in the participant
+list.
+
+~~~ tls
+struct {
+  IdentifierUri client_pseudonym;
+  IdentifierUri user_pseudonym;
+  opaque signature_public_key;
+} PseudonymousCredentialTBS
+
+struct {
+  /* SignWithLabel(., "PseudonymousCredentialTBS",
+    PseudonymousCredentialTBS) */
+  opaque pseudonymous_credential_signature<V>;
+  Credential client_credential;
+} IdentityLinkTBE
+~~~
+
+The `identity_link_ciphertext` is created by encrypting the IdentityLinkTBE,
+which contains the client's real credential, as well as a signature over the
+PseudonymousCredentialTBS using the client credential's signature key.
+
+The `identity_link_key` used for encryption is unique per pseudonymous credential.
+It is derived from the client's `connection_key`.
+
+~~~ tls
+identity_link_key = ExpandWithLabel(connection_key,
+  "IdentityLinkKey", PseudonymousCredentialTBS, KDF.Nh)
+~~~
+
+See {{connection-keys}} for more details on `connection_key`s.
+
+Pseudonyms are specific to each room and client since the `identity_link_key` is
+unique per pseudonymous credential and pseudonymous credentials are meant to be
+used only once (i.e. in one room). Re-use of pseudonymous credential can occur
+if KeyPackages are re-used. See {{pseudonymous-keypackages}}.
+
+## Identity link keys
+
+To allow all participants of a room's underlying group to link each pseudonym
+with a real identity, all identity link keys must be available to all
+participants, including new joiners. To that end, and to ensure agreement among
+group members, the `identity_link_keys` are stored encrypted in the
+IdentityLinkKeysExtension. The MLS group underpinning an MMR MUST contain an
+IdentityLinkKeysExtension.
+
+~~~ tls
+struct {
+  opaque ciphertext<V>;
+} EncryptedIdentityLinkKey
+
+struct {
+  optional<EncryptedIdentityLinkKey> identity_link_key_ciphertexts<V>;
+} IdentityLinkKeysExtension
+~~~
+
+The `identity_link_key_ciphertexts` vector has as many entries as the MLS group
+has leaves and it contains the encrypted `identity_link_key` of each group
+member at the leaf index of that member.
+
+The individual `identity_link_keys` of the participants are encrypted using the
+group's `identity_link_wrapper_key` (see {{identity-link-wrapper-key}}).
+
+When members are removed from the group, the corresponding entries in the
+IdentityLinkKeysExtension MUST be removed as well.
+
+When members are added to the group via a room state update (see
+{{update-room-state}}), the sender MUST use the SafeAAD mechanism described in
+{{!I-D.ietf-mls-extensions}} to include an AddAAD.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-mimi-protocol.md
```diff
@@ -1754,6 +1754,192 @@ struct {
...
+
+- `user_pseudonym`: The pseudonym of the client's user in this group
+- `client_pseudonym`: The pseudonym of the client identified by this credential
+- `signature_public_key`: The signature public key used to authenticate MLS
+  messages
+- `identity_link_ciphertext`: A ciphertext containing a credential with the
+  clients real identifier
+
+
+In each room, the `user_pseudonym` of a client MUST be the same across all
```
rohanmahy, line 1798:
```suggestion
In any given room, the `user_pseudonym` of a client MUST be the same across all
```

draft-ietf-mimi-protocol.md
```diff
@@ -1754,6 +1754,192 @@ struct {
...
+~~~
+
+See {{connection-keys}} for more details on `connection_key`s.
+
+Pseudonyms are specific to each room and client since the `identity_link_key` is
+unique per pseudonymous credential and pseudonymous credentials are meant to be
+used only once (i.e. in one room). Re-use of pseudonymous credential can occur
+if KeyPackages are re-used. See {{pseudonymous-keypackages}}.
+
+## Identity link keys
```
rohanmahy, line 1836: shouldn't this section heading move to before line 1821?

draft-ietf-mimi-protocol.md
```diff
@@ -1754,6 +1754,192 @@ struct {
...
+  opaque ciphertext<V>;
+} EncryptedIdentityLinkKey
+
+struct {
+  optional<EncryptedIdentityLinkKey> identity_link_key_ciphertexts<V>;
+} IdentityLinkKeysExtension
+~~~
+
+The `identity_link_key_ciphertexts` vector has as many entries as the MLS group
+has leaves and it contains the encrypted `identity_link_key` of each group
```
rohanmahy, line 1856: what happens with blank leaves. is the `identity_link_key` an empty <V>?

draft-ietf-mimi-protocol.md
```diff
@@ -1754,6 +1754,192 @@ struct {
...
+of the Add proposals in the commit adding the new members.
+
+## Identity link wrapper key
+
+TODO: The concept of an identity link wrapper key will be replaced by the
+TreeWrap protocol, which is currently still work in progress.
+
+The identity link wrapper key protects the `identity_link_key`s of a group while
+they are stored on the server. The `identity_link_wrapper_key` is sampled
+randomly when the group is created. It is static throughout the lifetime of the
```
rohanmahy, line 1886: If the key is static, then a former member (even one who was banned) could collude with the hub to decrypt the real identities of members who joined well after the former member left.

draft-ietf-mimi-protocol.md
```diff
@@ -1754,6 +1754,192 @@ struct {
...
+During connection establishment, users MUST exchange connection keys in a
+mutually authenticated way and in such a way that only the two users learn the
+key.
+
+When a user wants to add another user to an MMR, it fetches KeyPackages with
+PseudonymousCredentials that share a user pseudonym (see
+{{pseudonymous-keypackages}}). The adder then derives the `identity_link_key`
+for all KeyPackages and re-encrypts those keys under the MMR's
+`identity_link_wrapper_key`. When adding the user to the MMR, it includes the
+AddAAD as described in {{identity-link-keys}}.
```
rohanmahy, line 1925: I didn't really understand this paragraph. 
```suggestion
When a user wants to add another user to an MMR, it fetches one KeyPackage for each of that user's clients, such that each of the clients keypackages has a PseudonymousCredential that shares a single user pseudonym (see
{{pseudonymous-keypackages}}). The adder then derives the `identity_link_key`
for all the added client's KeyPackages and re-encrypts those keys under the MMR's
`identity_link_wrapper_key`. When adding the user to the MMR, it includes the
AddAAD as described in {{identity-link-keys}}.
```

draft-ietf-mimi-protocol.md
```diff
@@ -1754,6 +1754,192 @@ struct {
...
+
+~~~ tls
+struct {
+  EncryptedIdentityLinkKey identity_link_key_ciphertextexts<V>;
+} AddAAD
+~~~
+
+Each `identity_link_key_ciphertext` corresponds to an added group member. The
+order of ciphertexts in the `identity_link_ciphertexts` MUST be the same as that
+of the Add proposals in the commit adding the new members.
```
rohanmahy, line 1877: I'm guessing using an AddAAD is so external joins work? For adds originated by a member, they could use GCE or some similar mechanism to update the identity_link_key_ciphertext directly, right?

rohanmahy (commented):
Overall, there are really two mechanisms:

A) to provide a pseudonym, encrypting the real credential from the Hub, and so another member with the real credential can verify the real credential and the pseudonym was signed by the same signature key.

B) key management for distributing keys, eventually allowing members to decrypt the real credentials.

----
Some comments and questions:
Does this work with external joiners or not? That seemed ambiguous.

IMO there is not enough explanation of why a particular technical choice was selected (ex: AddAAD vs some other mechanism). This should be either motivated in the main text or in the (absent) security considerations.

It would be useful to walk the reader through the process to follow roughly the steps a client executes each join method (for example, a member creating a group, a member adding a new user to a group, a user adding themself via external commit, a user processing a welcome, and a member validating a commit).

Is there anything the MIMI hub or other providers need to do to find KeyPackages with suitable properties?

That the key for protecting the individual keys is shared for the whole group lifetime seems unsatisfactory.

draft-ietf-mimi-protocol.md
```diff
@@ -1754,6 +1754,192 @@ struct {
...
+  /* SignWithLabel(., "PseudonymousCredentialTBS",
+    PseudonymousCredentialTBS) */
+  opaque pseudonymous_credential_signature<V>;
+  Credential client_credential;
+} IdentityLinkTBE
+~~~
+
+The `identity_link_ciphertext` is created by encrypting the IdentityLinkTBE,
+which contains the client's real credential, as well as a signature over the
+PseudonymousCredentialTBS using the client credential's signature key.
```
rohanmahy, line 1819:
```suggestion
The `identity_link_ciphertext` is created by encrypting the IdentityLinkTBE.
The IdentityLinkTBE contains the client's real credential, and a signature over the
PseudonymousCredentialTBS signed with the client credential's `signature_public_key`.
```

rohanmahy:
After the MIMI interim, I think I like the idea of accepting this PR up to line 1821 and adding a note that the key exchange is going to be addressed separately. Then we can try a few keying solutions and see how will they fit our requirements (either for credentials only, or also for sensitive data in the GroupContext, like room metadata).

draft-ietf-mimi-protocol.md
```diff
@@ -1754,6 +1754,192 @@ struct {
...
+state: the participant list and the credentials of the room's underlying MLS
+group. In an MMR, the real identifiers of clients and users are replaced by
+pseudonyms in the shape of random UUIDs qualified with the domain of the user's
+provider.
+
+In MMRs, all leaves of the underlying group MUST contain PseudonymousCredentials.
+
+~~~ tls
+struct {
+  IdentifierUri client_pseudonym;
```
rohanmahy, line 1783: Is it really necessary to have a client_pseudonym? Couldn't we use a thumbprint of the signature key? Alternatively, one of the possible forms of a MIMI client identifier is to use `mimi://example.com/d/alice-smith/0003`, such that the d is for deice and when there are two more path segments after the "d", the first is a "user" identifier, and the second is a client discriminator.

rohanmahy:
One other issue that occurred to me is that the Hub could enforce some membership changes based on a member giving the pseudonym a role, but Preauthorization would be useless since the credentials are all or nothing. Selective disclosure or ZKPs in credentials would be more flexible.

draft-ietf-mimi-protocol.md
kkohbrok: The signature key isn't necessarily static, but maybe that's okay. Sure, we could also combine user and client pseudonym into a single identifier if you'd prefer that. No strong opinion here.

draft-ietf-mimi-protocol.md
kkohbrok: Yes, that's why it's `optional`. We could save those `None` bytes and pack the `identity_link_keys` tight, but I thought lining them up with the leaf indices would be more intuitive.

draft-ietf-mimi-protocol.md
kkohbrok: Yes. Somehow I thought the `identity_link_ciphertexts` would be stored outside the group, but storing them in the group context is probably better. I guess there's no reason not to use AppSync here. What do you think?

draft-ietf-mimi-protocol.md
kkohbrok: Yes, that's not ideal. We're working on a mechanism that gives us nice FS and PCS guarantees that line up with those we get from MLS.

kkohbrok (commented):
Thanks for the review! I've addressed a few of the issues and tried to answer some of your questions. Please take a look.

kkohbrok:
After a quick discussion OOB, I've cut the key management portion of the PR and replaced it with a TODO. I hope this reduces the PR to the portion we have agreement on s.t. we can merge before cutting the next draft version.

rohanmahy (approved):

PR #111, opened 2025-03-16
Create new proxyDownload endpoint

rohanmahy:
This replaces #75 

It adds Oblivious HTTP between the downloading provider and the Hub provider.

### DIFF
--- a/draft-ietf-mimi-protocol.md
+++ b/draft-ietf-mimi-protocol.md
@@ -752,7 +752,9 @@ GET /.well-known/mimi-protocol-directory
   "identifierQuery":
      "https://mimi.example.com/v1/identifierQuery/{domain}",
   "reportAbuse":
-     "https://mimi.example.com/v1/reportAbuse/{roomId}"
+     "https://mimi.example.com/v1/reportAbuse/{roomId}",
+  "proxyDownload"
+     "https://mimi.example.com/v1/proxyDownload/{downloadUrl}"
 }
 ~~~
 
@@ -1825,6 +1827,44 @@ struct {
 ~~~
 
 There is no response body. The response code only indicates if the abuse report was accepted, not if any specific automated or human action was taken.
+
+
+## Download Files
+
+IM systems make extensive use of inline images, videos, and sounds, and often include attached files, all of which will be referred to as "assets" in this section. Assets are stored (encrypted) on various external servers, and typically uploaded and fetched using HTTP {{!RFC9110}} protected with TLS {{RFC8446}}.
+
+Broadly, two approaches are used for storage of assets in federated IM systems.
+In the most common approach, the assets are uploaded to the local system of the uploader.
+In other systems, the assets are uploaded to the system hosting the equivalent of a MIMI room.
+MIMI supports both of these approaches.
+
+The `proxyDownload` endpoint defined in this section is used to prevent the provider hosting the asset from learning about the IP addresses, number, and online status of clients associated with another provider, when that client downloads an asset/attachment in a specific room.
+
+The client consults the intersection of the room policy and its local policy to determine if and how to upload assets, where to upload them, and with which credentials. The details of the upload are out of scope of this document.
+
+MIMI room policy {{!I-D.ietf-mimi-room-policy}} includes an asset policy that specifies which domain name to use for asset download for each provider's domain of a potential asset sender.
+
+When a receiving client receives a reference to an external asset sent by a participant from another provider, the receiving client downloads the asset over a proxy on the hub, protected using Oblivious HTTP (OHTTP) {{!RFC9458}}.
+The receiving client MUST verify that the download URL host part is associated with the asset server domain specified for the sender's identity.
+The receiving client uses its local provider's OHTTP Relay; the hub provider is the OHTTP Gateway; and the `proxyDownload` endpoint is the OHTTP Target Resource on the Hub.
+The first time the receiving client accesses the `proxyDownload` endpoint, it fetches the MIMI directory of the Hub provider for `proxyDownload` endpoint.
+It then uses an HTTP GET (sent over OHTTP using the local provider relay and the hub's gateway) to the `proxyDownload` endpoint as specified in the directory with the asset's `downloadUrl` of the asset.
+
+> The `downloadUrl` would be the `uri` field in an `ExternalPart` in a MIMI content {{!I-D.ietf-mimi-content}} message.
+
+~~~
+GET /proxyDownload/{downloadUrl}
+~~~
+
+The Hub MUST verify that the download URL host part is associated with the asset server domain for one of its peer providers.
+If the request succeeds, the response body contains the contents of the
+downloaded URL.
+The Hub is encouraged to cache the resources it downloads.
+
+In terms of the privacy of this mechanism, the Hub only sees that a specific provider requested an (opaque to the Hub) URL from another provider. It does not know the room or the specific client.
+The provider of the uploader sees requests for an (encrypted) asset coming from the Hub provider. When caching is employed, the provider will only see periodic requests.
+The receiver's provider will see one request to the Hub, among presumably many requests. It has no information about the sender, URL, or room associated with that request.
+
 
 # Minimal metadata rooms
 


draft-ietf-mimi-protocol.md
```diff
@@ -1826,6 +1828,44 @@ struct {
...
+## Download Files
+
+IM systems make extensive use of inline images, videos, and sounds, and often include attached files, all of which will be referred to as "assets" in this section. Assets are stored (encrypted) on various external servers, and typically uploaded and fetched using HTTP {{!RFC9110}} protected with TLS {{RFC8446}}.
+
+Broadly, two approaches are used for storage of assets in federated IM systems.
+In the most common approach, the assets are uploaded to the local system of the uploader.
+In other systems, the assets are uploaded to the system hosting the equivalent of a MIMI room.
+MIMI supports both of these approaches.
+
+The `proxyDownload` endpoint defined in this section is used to prevent the provider hosting the asset from learning about the IP addresses, number, and online status of clients associated with another provider, when that client downloads an asset/attachment in a specific room.
```
martinthomson, line 1841: I think that you are making stronger claims than you can really justify here.  In particular, I don't think that you can claim that the server doesn't learn about the number of clients.

draft-ietf-mimi-protocol.md
```diff
@@ -1826,6 +1828,44 @@ struct {
...
+
+The client consults the intersection of the room policy and its local policy to determine if and how to upload assets, where to upload them, and with which credentials. The details of the upload are out of scope of this document.
```
martinthomson, line 1843:
```suggestion
The client consults the intersection of the room policy and its local policy to determine if and how to upload assets, where to upload them, and with which credentials. The details of the upload process are out of scope of this document.
```

draft-ietf-mimi-protocol.md
```diff
@@ -1826,6 +1828,44 @@ struct {
...
+
+MIMI room policy {{!I-D.ietf-mimi-room-policy}} includes an asset policy that specifies which domain name to use for asset download for each provider's domain of a potential asset sender.
+
+When a receiving client receives a reference to an external asset sent by a participant from another provider, the receiving client downloads the asset over a proxy on the hub, protected using Oblivious HTTP (OHTTP) {{!RFC9458}}.
+The receiving client MUST verify that the download URL host part is associated with the asset server domain specified for the sender's identity.
+The receiving client uses its local provider's OHTTP Relay; the hub provider is the OHTTP Gateway; and the `proxyDownload` endpoint is the OHTTP Target Resource on the Hub.
+The first time the receiving client accesses the `proxyDownload` endpoint, it fetches the MIMI directory of the Hub provider for `proxyDownload` endpoint.
```
martinthomson, line 1850: This probably needs to specify where the client gets the hub's configuration.  The key configuration that the client uses is the configuration of the hub.

The local provider also needs to be able to setup relay endpoints that point to different hubs.

draft-ietf-mimi-protocol.md
```diff
@@ -1826,6 +1828,44 @@ struct {
...
+It then uses an HTTP GET (sent over OHTTP using the local provider relay and the hub's gateway) to the `proxyDownload` endpoint as specified in the directory with the asset's `downloadUrl` of the asset.
```
martinthomson, line 1851: You should draw a picture for this.

draft-ietf-mimi-protocol.md
```diff
@@ -1826,6 +1828,44 @@ struct {
...
+
+> The `downloadUrl` would be the `uri` field in an `ExternalPart` in a MIMI content {{!I-D.ietf-mimi-content}} message.
```
martinthomson, line 1853: I don't understand this bit.  Why do you have it quoted "> ..."?

More concretely, why are you putting the download URL on the end of /proxyDownload/ ?  That makes no sense to me.

Presumably, the client is going to make a request to the relay (POST with an encrypted BHTTP payload).  That payload will include a GET request to the URL that was attached to a message.

draft-ietf-mimi-protocol.md
```diff
@@ -1826,6 +1828,44 @@ struct {
...
+
+~~~
+GET /proxyDownload/{downloadUrl}
+~~~
+
+The Hub MUST verify that the download URL host part is associated with the asset server domain for one of its peer providers.
```
martinthomson, line 1859: This probably needs to be more carefully specified.  There is a case here where the hub gateway (and therefore the relay) is per-room and another where the gateway is global to the hub.  You'll need to decide what works.  It seems to me like the only reason why the hub does this check is so that it doesn't get used as an open relay.  So you could make this a single gateway endpoint for the entire hub.  On the other hand, if this is setup per-room, then you will allow the hub to make finer-grained access control decisions, at the cost of more configuration complexity and lower privacy.

draft-ietf-mimi-protocol.md
```diff
@@ -1826,6 +1828,44 @@ struct {
...
+If the request succeeds, the response body contains the contents of the
+downloaded URL.
+The Hub is encouraged to cache the resources it downloads.
```
martinthomson, line 1862: {{RFC6919}}

I think that you probably want to explain:

1. That the hub can cache resources.
2. What the consequences of that caching would be.

draft-ietf-mimi-protocol.md
```diff
@@ -1826,6 +1828,44 @@ struct {
...
+
+In terms of the privacy of this mechanism, the Hub only sees that a specific provider requested an (opaque to the Hub) URL from another provider. It does not know the room or the specific client.
```
martinthomson, line 1864: You probably need a more complete analysis of the privacy properties of this system.  Including taking the position that the hub is actively looking to undermine the privacy of room participants.

Consider the case where the hub selectively forwards messages to participants, for instance, by staggering delivery.  If clients are online, they might hit the gateway immediately after receiving a URL.  In that way, the hub knows both the URL that was in the message, and which client initiated that download.  It also learns how large the asset is.

This is likely the best that we can expect to achieve here, in that the hub already has this sort of information available to it.  Though it does create a covert channel from the message originator to the hub, which can be attached to individual recipients.

draft-ietf-mimi-protocol.md
```diff
@@ -1826,6 +1828,44 @@ struct {
...
+The provider of the uploader sees requests for an (encrypted) asset coming from the Hub provider. When caching is employed, the provider will only see periodic requests.
```
martinthomson, line 1865: "the provider of the uploader" probably needs a clear term; it's a little clunky.

draft-ietf-mimi-protocol.md
```diff
@@ -1826,6 +1828,44 @@ struct {
...
+The receiver's provider will see one request to the Hub, among presumably many requests. It has no information about the sender, URL, or room associated with that request.
```
martinthomson, line 1866: This is through the relay it provides, yes?  Maybe say so.

mar-v-in:
Previously, the mimi-protocol was mostly about the protocol between user's provider and hub and leaves the exact protocol between client and provider out of scope. The use of OHTTP adds specific requirements to the client in the whole design. 

I think stating that the Hub must provide an HTTP proxy with specific properties that is made available to the provider should be sufficient for the purpose of mimi-protocol: it allows for the client to use any protocol of their choice to instruct their provider to download the file via the hub's HTTP proxy.

mar-v-in:
I also think the protocol laid out here allows a provider server to use a hub server (that one of their user is using) as a proxy to any remote server by setting that remote server's domain (temporarily) as their asset server domain.

I also probably missed how the hub server is supposed to discover the asset server domains of its members: it's not in the directory and even if it was, the hub also can't know which provider's directory to query when requested to download a file.

draft-ietf-mimi-protocol.md
```diff
@@ -1826,6 +1828,44 @@ struct {
 
 There is no response body. The response code only indicates if the abuse report was accepted, not if any specific automated or human action was taken.
 
+
+## Download Files
+
+IM systems make extensive use of inline images, videos, and sounds, and often include attached files, all of which will be referred to as "assets" in this section. Assets are stored (encrypted) on various external servers, and typically uploaded and fetched using HTTP {{!RFC9110}} protected with TLS {{RFC8446}}.
```
raphaelrobert, line 1834: So far we have been calling them "attachments", including in draft-ietf-mimi-content. We should not introduce a different yet synonymous term.

draft-ietf-mimi-protocol.md
```diff
@@ -1826,6 +1828,44 @@ struct {
...
+
+Broadly, two approaches are used for storage of assets in federated IM systems.
+In the most common approach, the assets are uploaded to the local system of the uploader.
+In other systems, the assets are uploaded to the system hosting the equivalent of a MIMI room.
+MIMI supports both of these approaches.
```
raphaelrobert, line 1839: Do we have consensus on the fact that we want to support both approaches? I think in past discussions we wanted to focus on the hub-only approach.

rohanmahy: My recollection was that webex and teams supported their equivalent of the system for storage and we wanted to stay agnostic. In any case, the technical solution that restricts the downloads to come from one of the participants' local attachment servers allows any of them to delegate this URL to the hub server.

draft-ietf-mimi-protocol.md
rohanmahy: ok. I'll update.

bifurcation: I honestly like "assets" a little better, and I think it's slightly more commonly used in practice.

draft-ietf-mimi-protocol.md
bifurcation: AFAIK, neither Webex nor Teams is federated in any sense that would make a good analogy here.  The hub-only approach does seem simpler, though (a) it increases the cost to be a hub and (b) could impede hub transition later.  Given those drawbacks, having some flexibility here seem good.

draft-ietf-mimi-protocol.md
bifurcation: Maybe you could if the hub cached assets.  That could be a compromise between the hub-only approach and the flexible approach.

raphaelrobert:
As discussed during the ietf122 mimi session, I propose to simplify this by making the hub squarely responsible for anything related to attachments:
 - the hub hosts attachments
 - the hub can enforce quotas
 
 This makes for a far easier architecture and a more consistent UX.

mar-v-in:
In the end in any approach the file has to travel the CSSSC path and given we use HTTPS for message transfer on the S-to-S links, it feels only obvious to use it for the file transfer as well. So the question we're discussing effectively is if the transfer from sending server to hub is via GET or PUT. The architecture doesn't change by a lot.

Also, the hub in practice won't be able to completely enforce limits, simply because any file can likely be uploaded to an external service and displayed as a link that recipients can access using an external software.

draft-ietf-mimi-protocol.md
rohanmahy: @martinthomson  Yes, that was overstating things. 

The hub can't learn the number of participants from a provider B that are **in a specific room** if that provider has participants in several rooms on that hub. The hub only learns the total number of asset requests coming from the provider's OHTTP relay.  

Are you ok with some rephrasing of that?

rohanmahy:
> As discussed during the ietf122 mimi session, I propose to simplify this by making the hub squarely responsible for anything related to attachments:
> 
>     * the hub hosts attachments
>     * the hub can enforce quotas

I think we can expect near zero adoption from any major provider in that case. What's the motivation for a large provider to act as a hub for other clients?

draft-ietf-mimi-protocol.md
martinthomson: Yeah, the hub learns the number of requests from each provider, which is <= the number of participants, I expect (unless participants request content multiple times, which I guess is also possible).

The hosting entity would learn that someone has requested the content.

raphaelrobert:
> I think we can expect near zero adoption from any major provider in that case. What's the motivation for a large provider to act as a hub for other clients?

You claimed that at IETF122 already, but it would be good to back that claim with some concrete evidence.

I don't understand why large providers would be treated any differently than small ones. The provider who becomes the hub is decided at room creation. Rooms hosted by a specific provider can only be created by users of that provider. In that moment, a provider can decide

 a) if federating with other users is allowed, and
 b) if so, what the quota for the room is (either in total or per user), and
 c) what the retention time is.

I don't understand why there would be a bias towards large providers. And even if there was, providers remain in control of the cost.

rohanmahy:
> > I think we can expect near zero adoption from any major provider in that case. What's the motivation for a large provider to act as a hub for other clients?
> 
> You claimed that at IETF122 already, but it would be good to back that claim with some concrete evidence.

The cost of storing assets is one of the most expensive parts of a messaging service, assets need to be stored for a longer amount of time and brings a host of other liabilities.

If hub A has some user from provider B, we expect provider downtime to be fairly short (minutes or hours) with rare cases of a couple of days. In most cases, messages sent to provider B for participants can be cleaned up within minutes. By contrast the expectation is that assets/attachments are stored for a much longer amount of time, in the 30 days to 1 year time frame. They are also 3-4 orders of magnitude larger than typical messages. 

We have no proposed mechanism yet to allow a participant from a non-hub provider to upload assets. We know that the balance of authentication, privacy, and policy enforcement is a very hard balance and it is more challenging due to the multiparty nature of this flow.

Next we have legal liability for assets. If someone reports an asset, there are a variety of legal issues in most jurisdictions. For example for CSAM, the asset needs to be quarantined. In many jurisdictions it can't even be viewed by regular administrative staff without breaking the law. It may need to be kept as evidence for years. These are all accepted costs and risks for the hub's own users, but accepting these costs and risks for others is a disproportionate burden for assets than for messages.
 
> I don't understand why large providers would be treated any differently than small ones. The provider who becomes the hub is decided at room creation. Rooms hosted by a specific provider can only be created by users of that provider. In that moment, a provider can decide

They are not treated any differently, but all the large providers today are a free service to their users. Any providers who are free (large or small) do not get additional revenue from non-hub users. For text messaging the cost of hosting a room with several non-hub users is relatively low, but much higher if assets need to be supported.

I could easily image some small or medium sized providers that want to offer a paid service where everything is stored on the hub in certain rooms, but absent that there is no motivation to change the default modus operandi of messaging services.

Finally, in our charter we said we were going to try to have a light touch and not design a client to server protocol, nor dictate too much. All commercial messaging services that I know about already have a mechanism to store assets on the client's provider. This seems a pretty big departure for a questionable philosophical argument.

> a) if federating with other users is allowed, and b) if so, what the quota for the room is (either in total or per user), and c) what the retention time is.
> 
> I don't understand why there would be a bias towards large providers. And even if there was, providers remain in control of the cost.
I'm not sure I understand. For MIMI to be successful, it needs to work for both large and small providers. If no large providers can deploy it (or no small providers for that matter), then we have failed. That's not a bias, it is the recognition that a healthy ecosystem needs to contain both.

raphaelrobert:
Thanks for the added context. I feel we are going in circles because I have a hard time discerning what's an assumption in the above and what is an established fact.

To make some progress, I propose to zoom out a bit. I think we can agree on the following:
 a. The design philosophy has been to empower the hub to be the arbiter and enforcer of policies wherever possible
 b. There are concerns about the cost of hosting assets on the hub

We have discussed two options so far:
 1. Dismiss b. so that we remain true to a. (Raphael and others)
 2. Make an exception to a. to account for b. (Rohan)

Let's consider some middle ground, i.e., option 3. Because of the CSSSC architecture, attachments have to transit via the hub anyway. Rather than making the hub just a proxy, we could leave it to the hub whether it wants to cache attachments or not. The hub provider could decide whether it wants the cheapest option (where other providers exclusively store the attachments of their respective users) or whether it prefers to shoulder the hosting costs and provide better attachment availability. In both cases, the hub can still enforce policies regarding assets (e.g. quotas).

In a scenario with client A on a.com, client B on b.com, and hub H, the flows would look as follows:

**Upload**

 - Client A uploads an attachment to a.com
 - Client A advertises the attachment to the group via the hub by sharing the attachment id
 - Hub H can see (part of) that message and decide whether it wants to a) allow the upload and b) cache the attachment
 - In the case of b), the hub H downloads the attachment from a.com

**Download**

 - Client B has learned about a new attachment
 - Client B asks its provider b.com to download the attachment from the hub H by providing its id
 - Depending on the hub's policy, it either serves the attachment from its cache or proxies the request to a.com

All we need to do for this is to describe the API to upload/download attachments at the hub.

Splow95 (approved):

draft-ietf-mimi-protocol.md
rohanmahy: added text to explain that this is a component of the GroupContext, so it is set per room.

draft-ietf-mimi-protocol.md
rohanmahy: done

draft-ietf-mimi-protocol.md
rohanmahy: I outlined three approaches and we can discuss in Madrid

rohanmahy:
Merging so we have something to discuss, but not resolving the relevant issue.

PR #113, opened 2025-03-18
Fix the name of the key in SignWithLabel in GroupInfoResponse

rohanmahy:


bifurcation (changes requested):
This is not a "2-time pad" (terrible phrase; almost never a real thing), this is bounding the replayability of the signature.  You should either:

1. Put a nonce in the request, which the response signature must include
2. Put a timestamp in the response, which the requester compares to their local time

Clearly (1) would be preferable, because it doesn't require time sync.

draft-ietf-mimi-protocol.md
```diff
@@ -1458,7 +1458,9 @@ struct {
 
 If successful, the response body contains the GroupInfo and a way
 to get the ratchet_tree, both encrypted with the `groupInfoPublcKey`
-passed in the request.
+passed in the request. The response contains a unique random
+`group_info_response_nonce` to prevent the Hub's signature key from being used
+repeatedly with the same inputs.
```
rohanmahy, line 1463:
```suggestion
passed in the request.
```

draft-ietf-mimi-protocol.md
```diff
@@ -1491,6 +1493,7 @@ struct {
       CipherSuite cipher_suite;
       ExternalSender hub_sender;
       HPKECiphertext encrypted_groupinfo_and_tree;
+      uint8 group_info_response_nonce[16];
```
rohanmahy, line 1496:
```suggestion
```

draft-ietf-mimi-protocol.md
```diff
@@ -1505,8 +1508,9 @@ struct {
           CipherSuite cipher_suite;
           ExternalSender hub_sender;
           HPKECiphertext encrypted_groupinfo_and_tree;
-          /* SignWithLabel(hub_sender, "GroupInfoResponseTBS", */
-          /*                GroupInfoResponseTBS) */
+          uint8 group_info_response_nonce[16];
```
rohanmahy, line 1511:
```suggestion
```

PR #118, opened 2025-04-25
Change franking_integrity to use signature

rohanmahy:
- Change franking_integrity to use signature
- create franking_agent GroupContext component to hold signature public key for validating franking_integrity_signature
- clarify relationship with MIMI content salt, sender_uri, and room_uri
- register all four app components with IANA
- acknowledge Felix Linker for Tamarin formal methods model
- regularize case conventions for franking-related fields


draft-ietf-mimi-protocol.md
```diff
@@ -1134,128 +1136,139 @@ Franking was popularized by Facebook and described in their whitepaper
...
 {
-  / RoomURI        / 1: "mimi://hub.example/r/Rl33FWLCYWOwxHrYnpWDQg",
-  / SenderUserURI  / 2: "mimi://b.example/u/alice"
+  / sender_uri / 1: "mimi://b.example/u/alice"
+  / room_uri   / 2: "mimi://hub.example/r/Rl33FWLCYWOwxHrYnpWDQg",
 }
 ~~~
 
-Note that this assertion does not vouch for the validity of these values,
+Note that these assertions does not vouch for the validity of these values,
```
rohanmahy, line 1191:
```suggestion
Note that these assertions do not vouch for the validity of these values,
```

draft-ietf-mimi-protocol.md
```diff
@@ -2168,18 +2183,52 @@ cost is offset by the simplicity of not having multiple policy enforcement point
...
 TBD.
 
->**TODO**: Present the franking mechanism to CFRG for review.
-
 
 # IANA Considerations
 
->**TODO**: Add registration of MIMI content format extension, and a SafeAAD
-component.
+This document registers the following two MLS application components per
```
rohanmahy, line 2189:
```suggestion
This document registers the following four MLS application components per
```

draft-ietf-mimi-protocol.md
```diff
@@ -2168,18 +2190,52 @@ cost is offset by the simplicity of not having multiple policy enforcement point
...
+- Value: TBD1
+- Name: frank_aad
+- Where: AD
+- Recommended: Y
+- Reference: RFCXXXX
+
+## franking_signature_key app component {#iana-franking-sig-key}
+
+- Value: TBD1
+- Name: frank_aad
```
rohanmahy, line 2210:
```suggestion
- Value: TBD2
- Name: franking_signature_key
```

draft-ietf-mimi-protocol.md
```diff
@@ -1134,128 +1136,146 @@ Franking was popularized by Facebook and described in their whitepaper
...
+When the Hub receives an acceptable application message with the `frank_aad`
+Safe AAD component and a valid sender identity, it calculates a server frank for
+the message, and a `franking_integrity_signature` as follows:
 
 ~~~
-context = senderURI || roomURI || acceptedTimestamp
-serverFrank = HMAC_SHA256(HUBkey, franking_tag || context )
-franking_integrity_check = HMAC_SHA256(franking_integrity_secret,
-  serverFrank || context)
+context = sender_uri || room_uri || accepted_timestamp
```
bifurcation, line 1230: Concatenation here is not good.  Risk of confusion between the two URIs.  Better to do this with something like a TLS-syntax or CBOR struct so that it's length-delimited.  Likewise for the other concatenations here.

draft-ietf-mimi-protocol.md
```diff
@@ -1134,128 +1136,146 @@ Franking was popularized by Facebook and described in their whitepaper
...
+MIMI franking relies on two new MLS application components. The first is the
+`frank_aad` Safe AAD component {{Section 4.9 of !I-D.ietf-mls-extensions}}. The
+second is the `franking_agent` GroupContext component {{Section 4.6 of
+!I-D.ietf-mls-extensions}}. The structure of `franking_agent` mirrors that of
+the `ExternerSender` struct described in {{Section 12.1.8.1 of !RFC9420}}. It
+contains a single signature key.
+
+~~~ tls
+struct {
+  uint8[32] franking_tag
```
bifurcation, line 1148: Perhaps this should be a vec, instead of being fixed-size forever?

draft-ietf-mimi-protocol.md
```diff
@@ -1134,128 +1136,146 @@ Franking was popularized by Facebook and described in their whitepaper
...
+} FrankAAD;
+
+FrankAAD frank_aad;
+
+struct {
+  SignaturePublicKey franking_signature_key;
+  Credential credential;
```
bifurcation, line 1155: I'm not thrilled that we're using Credential for a whole bunch of things that are not MLS clients.  But I don't have a great alternative in mind.

bifurcation (approved):
Couple of minor comments, overall LGTM.

draft-ietf-mimi-protocol.md
rohanmahy: The franking tag is strongly tied to the specific HMAC used in the franking algorithm, which uses HMAC_SHA256 exclusively right now. We would need to change a lot more if this becomes no longer prudent, at which point we can add a new MIMI protocol endpoint version or extension to switch to a new franking algorithm. This might work by defining a franking_tag_2 component for example.

PR #120, opened 2025-05-12
Make sure a new version of the hub's own invalidated pending proposals is stapled to an external commit

rohanmahy:


bifurcation (changes requested):
This needs more prose to describe what you're doing here.  It seems like the workflow you have in mind is:

* Hub generates some proposals, cues those up
* Someone comes along and does an external Commit
* Unfortunately, that external Commit can't include the hub's proposals
* So the Hub regenerates the proposals in the new epoch and distributes them

Assuming I have that right, I'm not clear on why you're sending the new proposals out like this.  (Same with the `moreProposals` field above.)  Is the expectation that a follower server or client is keeping a copy of the hub's proposal queue?   If that's the case, shouldn't we just send the whole proposal queue in these cases?  Wouldn't it be better just to send the proposal queue to a committer at the time they try to commit?

rohanmahy:
> This needs more prose to describe what you're doing here. It seems like the workflow you have in mind is:
> 
>     * Hub generates some proposals, cues those up
> 
>     * Someone comes along and does an external Commit
> 
>     * Unfortunately, that external Commit can't include the hub's proposals
> 
>     * So the Hub regenerates the proposals in the new epoch and distributes them
> 
> 
> Assuming I have that right, 

you've got that right.

>I'm not clear on why you're sending the new proposals out like this. (Same with the `moreProposals` field above.) Is the expectation that a follower server or client is keeping a copy of the hub's proposal queue? If that's the case, shouldn't we just send the whole proposal queue in these cases? Wouldn't it be better just to send the proposal queue to a committer at the time they try to commit?

Unless those proposals are legal inside an external commit (SelfRemove, AppDataUpdate, AppEphemeral), the client cannot just get a copy of the proposal queue, because their external commit would fail validation.

PR #127, opened 2025-07-11
Include valid proposals sent during the current epoch with a GroupInfo response

rohanmahy:
The MLS group for a room is in epoch <n>. 

Valid SelfRemove proposals are received from a member, or Valid AppDataUpdate or AppEphemeral external proposals are received from an external_sender.

The hub needs to provide these valid pending proposals which would be included in an external commit.

### DIFF
--- a/draft-ietf-mimi-protocol.md
+++ b/draft-ietf-mimi-protocol.md
@@ -1514,9 +1514,9 @@ struct {
 } GroupInfoRequest;
 ~~~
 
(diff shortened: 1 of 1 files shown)


rohanmahy:
Already needed for SelfRemove, AppDataUpdate, and AppEphemeral proposals. Could potentially be any type if https://www.ietf.org/archive/id/draft-mahy-mls-ext-commit-pp-00.html is adopted.

Issue #129, opened 2025-09-12
frank does not contain signature algorithm


rohanmahy:
If the signature algorithm used in a room changes (for example if the MLS group is reinitialized with a different ciphersuite), there is no way for the Hub to validate the `franking_integrity_signature` other than keeping track of the time it was changed or trying to brute force validate all the algorithms it has ever used. 

consider adding the signature algorithm to the frank as follows:

```
struct {
  uint8 server_frank[32];
  uint8 signature_algorithm;
  opaque franking_integrity_signature<V>;
} Frank;
```

Issue #122, opened 2025-07-05
Do we need franking context to use a struct or separators?


rohanmahy:
Currently the message context used in franking is defined as the concatenation of 3 fields:

```
context = sender_uri || room_uri || accepted_timestamp
```

There is a potential that these could be manipulated so a specially crafted sender_uri and room_uri could cause one client to appear as a blank user

Actually, there is no requirement that context be unique. We just need the context to correctly match the context fields in the sent message.

Snipped from [MIMI Content Issue #61 ](https://github.com/ietf-wg-mimi/draft-ietf-mimi-content/issues/61):
"While you could have:
`mimi://example.com/u/alicemimi://example.com/r/clubhouse` plus `mimi://example.com/r/clubhouse` in one message AND
`mimi://example.com/u/alice` plus `mimi://example.com/r/clubhousemimi://example.com/r/clubhouse` in another message,
it would take a pretty dumb hub to allow such a room ID and an AS with very permissive sender URI names."

rohanmahy:
I did this initially using a the fixed-sized approach and it was sort of ugly. In addition, the client already needs to do a handful of other things with TLS PL. The keys it is verifying with will need to come from a library that supports TLS PL anyway to hold the public signature key of the franker and its Credential.

I updated the PR appropriately.

PR #130, opened 2025-09-12
More franking fixes

rohanmahy:
- Added an explicit (uint16) lengths for the sender URI and room URI in the franking context. Addresses #122 
- Added a `franking_signature_ciphersuite` to allow the hub to validate a `franking_integrity_signature` even after the MLS ciphersuite has changed in that room. Addresses #129

### DIFF
--- a/draft-ietf-mimi-protocol.md
+++ b/draft-ietf-mimi-protocol.md
@@ -1113,7 +1113,8 @@ enum {
 } SubmitResponseCode;
 
 struct {
-  uint8[32] server_frank;
+  uint8 server_frank[32];
+  uint16 franking_signature_ciphersuite;
   opaque franking_integrity_signature<V>;
 } Frank;
 
@@ -1187,7 +1188,8 @@ FrankingAgentData FrankingAgentUpdate;
 ~~~
 
 The signature algorithm of the `franking_signature_key` is the same as the
-signature scheme in the cipher suite of the MLS group.
+signature algorithm of cipher suite of the MLS group which corresponds to the
+room.
 
 FrankingAgentUpdate is the format of the `update` field inside the AppDataUpdate
 struct in an AppDataUpdate Proposal for the `franking_agent` component. The use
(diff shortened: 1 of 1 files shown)


draft-ietf-mimi-protocol.md
```diff
@@ -1113,7 +1115,8 @@ enum {
 } SubmitResponseCode;
 
 struct {
-  uint8[32] server_frank;
+  uint8 server_frank[32];
+  uint16 franking_signature_ciphersuite;
```
bifurcation: Do you want to have an explicit type for the ciphersuite?  Like [MLS does](https://rfcs.online/rfcs/rfc9420.html#section-17.1)

draft-ietf-mimi-protocol.md
```diff
@@ -1113,7 +1115,8 @@ enum {
 } SubmitResponseCode;
 
 struct {
-  uint8[32] server_frank;
+  uint8 server_frank[32];
```
bifurcation: This should probably be a vector instead of something syntactically fixed-length, to allow for cipher agility.

draft-ietf-mimi-protocol.md
```diff
@@ -1172,7 +1175,7 @@ contains a single signature key.
 
 ~~~ tls
 struct {
-  uint8[32] franking_tag
+  uint8 franking_tag[32];
```
bifurcation: Ditto comment about fixed-length.

bifurcation (approved):

draft-ietf-mimi-protocol.md
```diff
@@ -1254,11 +1262,25 @@ When the Hub receives an acceptable application message with the `frank_aad`
...
+~~~ tls
+struct {
+    IdentifierUri sender_uri;
+    IdentifierUri room_uri;
+    uint64 accepted_timestamp;
+} ServerFrankingContext;
+
+ServerFrankingContext context;
+
+server_frank = HMAC_SHA256(hub_key, (franking_tag || context) )
```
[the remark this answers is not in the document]
rohanmahy: Yes. The way to update the MAC is to update the entire franking mechanism.

